PR #7, opened 2017-09-07
Rename mr -> merge-request

zaquestion:
`lab mr` was originally added because cobra does not implicitly
support commands with `-`'s in them and it was believed that `hub pr`
and hub `pull-request` were aliases. It is now understood that these
are seperate commands. This change corrects the mistake by renaming the
mr command.

### DIFF
--- a/cmd/mergeRequest.go
+++ b/cmd/mergeRequest.go
@@ -26,7 +26,7 @@ const (
 
 // mrCmd represents the mr command
 var mrCmd = &cobra.Command{
-	Use:   "mr",
+	Use:   "merge-request",
 	Short: "Open Merge Request on GitLab",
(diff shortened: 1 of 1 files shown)


zaquestion:
Hub's `pull-request` is being deprecated. https://github.com/github/hub/issues/1536

Definitely want to support mr as described for `pr` in the linked issue. For completeness it's probably best to also support `merge-request` since hub intends to keep their equivalent around.

adamryman:
I would say creating an issue for creating the `pr`/`mr` sub commands would be good. Linking that issue. 

I vote for quickly adding the `mr create` subcommand asap so typing `merge-request` does not become too much of a habit.

PR #14, opened 2017-09-19
Add `lab mr` commands checkout, list, create

zaquestion:
Implementing these in an attempt to follow the ideas presented here:
https://github.com/github/hub/issues/1536

Honestly, not sure how I feel about these changes. Some considerations:
- `lab mr checkout` could be `lab checkout` with some special cases to
 determine if it should checkout the ref
- Aliasing `lab mr` and `lab merge-request`
- `lab mr list` feels better as `lab list mrs`

Following for edit
- `lab edit mr`
- `lab edit issue`

### DIFF
--- a/cmd/mergeRequest.go
+++ b/cmd/mergeRequest.go
@@ -1,141 +1,18 @@
 package cmd
 
 import (
-	"bytes"
-	"fmt"
-	"log"
-	"regexp"
-	"runtime"
-	"strconv"
-	"strings"
-	"text/template"
-
 	"github.com/spf13/cobra"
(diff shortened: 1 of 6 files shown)


bmess:
Given that `lab edit mr` and `lab edit list` can be two clearly different things, why not show that explicitly by having the sub commands own the noun:
clearly `lab mr edit` is only going to edit the `mr` command's target.  Explicit is good! 💯

zaquestion:
Well, pretty much everyone I've spoken with has been in favor of `lab mr <subcommand>` I'm going to bring this in as is.

Issue #21, opened 2017-10-06
lab mr checkout is checking out from the current branch


zaquestion:
ref is getting fetched to head instead of branch name. So the branch doesn't point to the ref and the checkout is incorrect. 
```
$ git mr checkout 587
From gitlab.com/phony
 * branch              refs/merge-requests/111/head -> FETCH_HEAD
   1234567..1234567             -> phony
Switched to branch 'phony'
```

PR #22, opened 2017-10-06
[#21] (mr checkout) properly fetch MR ref to branchname

zaquestion:


cmd/mrCheckout.go
```diff
@@ -39,7 +39,7 @@ var checkoutCmd = &cobra.Command{
 		}
 		branch := mrs[0].SourceBranch
 		mr := fmt.Sprintf("refs/merge-requests/%s/head", mrIDStr)
-		gitf := git.New("fetch", targetRemote, mr, fmt.Sprintf(":%s", branch))
+		gitf := git.New("fetch", targetRemote, fmt.Sprintf("%s:%s", mr, branch))
```
hasAdamr: Possibly add a comment that links back to the git documentation? Or which man page shows you how to do this?

adamryman: Possibly add a comment that links back to the git documentation? Or which man page shows you how to do this?

zaquestion: Good point, putting a link here as well:  https://docs.gitlab.com/ee/user/project/merge_requests/#checkout-merge-requests-locally

Issue #11, opened 2017-09-11
`lab clone` without url causes panic instead of printing help


hasLeland:
## What was expected?

When running `git clone` without specifying a url, I get a help message:

```
$ git clone
You must specify a repository to clone.

usage: git clone [<options>] [--] <repo> [<dir>]

    -v, --verbose         be more verbose
    -q, --quiet           be more quiet
...
...
```

## What happened instead?

When I ran `lab clone` without a url, `lab` panic'd:
```
$ lab clone
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x58350f]

goroutine 1 [running]:
github.com/zaquestion/lab/vendor/github.com/spf13/cobra.(*Command).ParseFlags(0x9c6a40, 0xc42000c310, 0x1, 0x1, 0x9c6a40, 0x9e69a0)
	/home/leland/projects/go-projects/src/github.com/zaquestion/lab/vendor/github.com/spf13/cobra/command.go:1267 +0x3f
github.com/zaquestion/lab/cmd.Execute()
	/home/leland/projects/go-projects/src/github.com/zaquestion/lab/cmd/root.go:81 +0x122
main.main()
	/home/leland/projects/go-projects/src/github.com/zaquestion/lab/main.go:6 +0x20
```

## How might this be fixed?

From what I can see in the `cmd/root.go` file, the problem is in [an unchecked slice of the `os.Args` slice on line 81.](https://github.com/zaquestion/lab/blob/f9c140fdce21cd24095bcdc58cd92bfc0fdda20f/cmd/root.go#L81) Directly above that line is a somewhat confusing comment which may or may not have anything to do with this issue, I had a hard time understanding it, though it may be of some use to whoever choses to fix this.

The change would be to check if `clone` has been provided without any additional arguments/parameters, and if it has, to print a help message in some way. The printing of a help message could be done by dropping down to gits help message or by printing `lab`s own custom help message.

zaquestion:
@hasLeland Thanks for reporting! I've put in a fix. This issue still needs tests to close.

bmess:
I apologize for beating a dead horse but I think this might be an elegant solution for an elegant tool :100:

bmess:
I believe this is fixed.  Tested with `0.5.0`. I closed the MR

PR #23, opened 2017-10-14
fixing issue #11 hopefully the cobra way

bmess:
I think using the Args argument might be cleaner.

### DIFF
--- a/cmd/clone.go
+++ b/cmd/clone.go
@@ -5,6 +5,7 @@ import (
 	"os"
 	"strings"
 
+	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"github.com/zaquestion/lab/internal/git"
 	"github.com/zaquestion/lab/internal/gitlab"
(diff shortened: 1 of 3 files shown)


bmess:
After updating the dependencies in cb3c88e and updating the message in 3baf542 I believe this is ready for review again

cmd/clone.go
```diff
@@ -18,6 +19,12 @@ var cloneCmd = &cobra.Command{
 	Long: `Clone supports these shorthands
 - repo
 - namespace/repo`,
+	Args: func(cmd *cobra.Command, args []string) error {
```
zaquestion: Can you implement something like [this](https://github.com/zaquestion/lab/blob/master/cmd/version.go#L15) here? This will allow us to pass the git clone output through. Ideally we'd want to ditch any lab/cobra output as well and make it look like we aren't involved.

zaquestion:
@bmess I've gone ahead and updated the dependencies for #25 and https://github.com/zaquestion/lab/commit/6538c02d3ef9ba9f936c9c35da5257db493996fa -- For whatever reason, this PR doesn't seem to be picking up the appropriate diff. Can you cherry-pick/reapply your changes from the latest master and re-open?

cmd/clone.go
bmess: done!

bmess:
The behavior appears to be fixed using `0.5.0` I'm going to close this

PR #39, opened 2017-11-09
Update to reduce calls to FindProject (hopefully)

adamryman:


adamryman:
Not actually useful. Was fun though.

Issue #30, opened 2017-10-27
Support issue command with options to create and list


zaquestion:
Similar to the `lab mr` command we want to be able to create and list issues for 1.0

Possible other subcommands
```
edit
comment
```

zaquestion:
Issue Create and MR Create tests were added in: https://github.com/zaquestion/lab/pull/55

Issue #29, opened 2017-10-27
Support custom issue and mr templates


zaquestion:
Ref: https://gitlab.com/help/user/project/description_templates.md#setting-a-default-template-for-issues-and-merge-requests

Useful for creating templates to `@mention` other people you work with on new issues or MRs.

PR #44, opened 2017-11-15
Integration Tests

zaquestion:
This PR adds a pattern for creating integration tests within the repo and adds tests for #11, #21, #29, and the list functionality in #30


jnv:
Is it good idea to have private key and token public? I understand those are for the testing account, but still...

You can encrypt SSH key file easily for Travis: https://docs.travis-ci.com/user/encrypting-files/

zaquestion:
@jnv Made them public so that they can run on pull requests to the repo. That restriction is documented here: https://docs.travis-ci.com/user/pull-requests#Pull-Requests-and-Security-Restrictions

They are for a test account and the test repo itself is on my gitlab account, so much of the attack surface is minimized since they can't push to it. Still, we'll see how it goes...

Issue #45, opened 2017-11-18
token asked for wrong domain


glensc:
```
$ lab remote -v
Enter default GitLab host (default: https://gitlab.com): https://gitlab.example.net 
Enter default GitLab user: glen
Enter default GitLab token (scope: api ):
Create here: https://gitlab.com/profile/personal_access_tokens[]
```

should say: https://gitlab.example.net/profile/personal_access_tokens

also, there's some usability issue, the cursor is placed in `[]`, i.e without newline. perhaps print that information before prompting user (swap two last lines)

also, the `stty echo` needs to be trurned off, not to show token when pasting it to terminal.

zaquestion:
Thanks for reporting!

zaquestion:
@glensc I've put in a fix, will keep this open for a bit longer to give you a chance to try it out first.

PR #46, opened 2017-11-18
[#45] (internal/gitlab) fix token asked for wrong domain + display issues

zaquestion:
- Use the user-defined GitLab host when printing the token key
- fix the cursor to prompt on the "Enter...token" line
- dont echo the token when its being entered


zaquestion:
Not sure how I feel about bring in a such a large dependency for this. Might change it latter, but its a good fix for now.

PR #53, opened 2017-12-02
Code coverage support with codecov integration

zaquestion:


zaquestion:
Looks like coverage is being generated and properly uploaded to codecov for the integration tests. https://codecov.io/gh/zaquestion/lab/commit/cd588abb1a852152c9422d788b3837a1fb3de2bf

Coverage is not getting created for `internal/gitlab` and `internal/git` from debugging it seems to be related to the use of the special `internal` package

PR #66, opened 2017-12-17
cmd: Add Issue show command

nkprince007:


cmd/issueShow.go
```diff
@@ -0,0 +1,100 @@
...
+
+	"github.com/spf13/cobra"
+	"github.com/xanzy/go-gitlab"
+	"github.com/zaquestion/lab/internal/git"
+	lab "github.com/zaquestion/lab/internal/gitlab"
+)
+
+var issueShowCmd = &cobra.Command{
+	Use:        "show [remote]",
+	ArgAliases: []string{"s", "d"},
```
zaquestion: Can you add a regular alias for `get`
```
Aliases: []string{"get"}
```

Also, consider dropping `"d"` as a short alias `"d"` makes me think "delete" over "describe".

cmd/issueShow.go
```diff
@@ -0,0 +1,100 @@
...
+			issue.TimeStats.HumanTotalTimeSpent)
+	}
+	if issue.DueDate != nil {
+		dueDate = time.Time(*issue.DueDate).String()
+	}
+	if issue.Assignee.Username != "" {
+		assignee = issue.Assignee.Username
+	}
+
+	fmt.Printf(`
```
zaquestion: This should be a template, but I'm tempted to say we can follow up on it.
Sample: https://github.com/zaquestion/lab/blob/master/cmd/issueCreate.go#L71

cmd/issueShow.go
```diff
@@ -0,0 +1,100 @@
...
+	Run: func(cmd *cobra.Command, args []string) {
+		issueNum := args[0]
+		if len(args) > 1 {
+			issueNum = args[1]
+		}
+		issueNumber, err := strconv.Atoi(issueNum)
+		if err != nil {
+			log.Fatal(err)
+		}
+		remote, _, err := parseArgsRemote(args)
```
zaquestion: The beginning of this function can just be
```
remote, issueNumber, err := parseArgsRemote(args)
```

cmd/issue.go
```diff
@@ -6,13 +6,30 @@ import (
...
+			return
+		}
+
+		issueShowCmd.Run(cmd, args)
 	},
 }
 
 func init() {
+	issueCmd.Flags().BoolP("list", "l", false, "list issues")
+	issueCmd.Flags().BoolP("create", "c", false, "create new issue")
```
zaquestion: Can we drop the create out of this one? I wanna think about it more.

PR #69, opened 2017-12-17
cmd: Add MR show command

nkprince007:


cmd/mr.go
```diff
@@ -7,13 +7,24 @@ import (
 // mrCmd represents the mr command
 var mrCmd = &cobra.Command{
 	Use:   "mr",
-	Short: "Work with merge requests",
-	Long:  ``,
+	Short: mrShowCmd.Short,
+	Long:  mrShowCmd.Long,
```
zaquestion: It seems like the Long and Short descriptions should stay how they are since the `mr` cmd has other functions under it. I realize I let this slip on the issue command, am intending to fix that before release. Potentially we could not in the long description that `lab mr 1` will do a `show`

Issue #48, opened 2017-11-26
Lab shouldn't store access credentials in .gitconfig


JoeNyland:
It's not uncommon for users to version control their dotfiles. A common file to version control is `~/.gitconfig`.

Lab currently stores and retrieves a users access token from `git config`. Having my Gitlab access token stored in `.gitconfig` means that I can no longer version control this file, which it important to me as this file contains shared configuration for all of my projects on my system.

If I understand the documentation for Lab currently, I _could_ add my access credentials to `.git/config` in each projects repository on my system, but this is a very tedious and cumbersome workaround to the issue.

[github/hub](https://github.com/github/hub) Helps it's users in this case by storing access tokens in an application specific config file `~/.config/hub`, which is a file that's less likely to be version controlled by users. I think Lab should use a similar solution, or at least provide it as an alternative credential store to the current `git config` system.

[github/hub](https://github.com/github/hub)s implementation is [here](https://github.com/github/hub/blob/cbd0189f60c71355f1903759c397a69cd3a0c861/github/config.go), if it helps.

zaquestion:
Thanks @JoeNyland, this seems like a fair assessment of current situation and I appreciate that you have specific usecases in mind. I've been talking about adding a config file for a few different things now. I'll move the global credentials in there when the time comes.

JoeNyland:
No worries! Ok, that sounds great! 🎉

zaquestion:
@JoeNyland Planning to implement these changes soon and I wanted to offer a possible workaround in the meantime. 

It looks like git 1.7.10+ supports sourcing other files for your .git config.
Ref: https://stackoverflow.com/questions/1557183/is-it-possible-to-include-a-file-in-your-gitconfig
I do something similar with my `.bashrc` by referencing a `.bashrc.local`.

This works for me using
```
$ git version
git version 2.11.0
lab version 0.5.2
```
`.gitconfig`
```
[user]
        name = Zaq? Wiedmann
        email = zaquestion@gmail.com
[include]
        path = .gitconfig.local
```
`.gitconfig.local`
```
[gitlab]
        host = https://gitlab.com
        user = lab-testing
```

nkprince007:
I think this has been solved. Closing it.

zaquestion:
Not quite, it'll go out in `0.8.0`

PR #70, opened 2017-12-18
[#48] (config) support lab.hcl config and move credentials into it.

zaquestion:


zaquestion:
This code wasn't tested to begin with. Its in the diff because the logic was moved.to a different file. Also its going to be removed when the legacy logic is ripped out (at least by `1.0`) and what remains will be tested. Going to merge this in as is, the existing tests definitely indicate the new config is working.

PR #71, opened 2017-12-18
Add `browse` sub command

nkprince007:


zaquestion:
I'm cool with dropping the committed vendor dir. Would like to see an updated way to install lab from source. Suggest writing a script that can curled into bash and using that. The build should set the `main.version` variable (like goreleaser) as well

cmd/issueBrowse_test.go
```diff
@@ -0,0 +1,19 @@
+package cmd
+
+import (
+	"testing"
+)
+
+func Test_issueBrowse(t *testing.T) {
```
zaquestion: You'll need some additional setup here to read the config since its usually done in main
```
	viper.SetConfigName("lab")
	viper.SetConfigType("hcl")
	viper.AddConfigPath(".")
	err = viper.ReadInConfig()
```

cmd/issueBrowse.go
```diff
@@ -0,0 +1,61 @@
...
+			log.Fatal(err)
+			return
+		}
+		if remote == "" {
+			remote = forkedFromRemote
+		}
+		rn, err := git.PathWithNameSpace(remote)
+		if err != nil {
+			log.Fatal(err)
+			return
```
zaquestion: `log.Fatal` will exit no need to return here.

internal/browser/browse.go
```diff
@@ -0,0 +1,27 @@
...
+
+	switch runtime.GOOS {
+	case "darwin":
+		cmd = "open"
+	case "linux", "freebsd", "openbsd":
+		cmd = "xdg-open" // wherever the X server is used
+	case "windows":
+		cmd = "cmd"
+		args = []string{"/c", "start"}
+	case "default":
```
zaquestion: Do you mean 
```
default:
    return errors.New("platform not supported")
```

nkprince007:
Wonder why codecov isn't reporting anything at all.

nkprince007:
Closing in favour of https://github.com/zaquestion/lab/pull/73

zaquestion:
@nkprince007 It actually updates the original comment. It threw me off as well.

PR #119, opened 2018-03-11
Add install to Makefile

adamryman:


Makefile
```diff
@@ -1,3 +1,11 @@
+install:
```
zaquestion: Maybe make `dep` a dependency of `install` and update the readme to just call that?

PR #120, opened 2018-03-12
[#104] Support setting the remote during `lab mr checkout`

### LINKED ISSUE #104: Support setting the remote during `lab mr checkout`
Flags:
-t, --track set checked out branch to track mr source remote, adds remote if needed

It probably makes since to detect if the mr source remote matches the `forkedFromRemote`, or `forkRemote` and to set tracking. Otherwise adding a new remote using the username scheme.

adamryman:
Works as follows:
```
lab mr checkout -t !mrID
```

When `-t` or `--track` is specified the following happens.

- Check for remote by the username if the MR author
    - If the remote does not exist, add it as `mr.Author.Username
- Instead of fetching to the configured branch, fetch to `refs/remotes/mr.Author.Username/mr.sourceBranch` (`fetchToRef`)
- Create a local branch starting from `fetchToRef`
- Checkout that branch

Test on the way.

Closes #104

### DIFF
--- a/cmd/mrCheckout.go
+++ b/cmd/mrCheckout.go
@@ -5,13 +5,20 @@ import (
 	"log"
 
 	"github.com/spf13/cobra"
+	"github.com/tcnksm/go-gitconfig"
 	"github.com/xanzy/go-gitlab"
 	"github.com/zaquestion/lab/internal/git"
 	lab "github.com/zaquestion/lab/internal/gitlab"
 )
 
-var (
+// mrCheckoutConfig holds configuration values for calls to lab mr checkout
+type mrCheckoutConfig struct {
 	branch string
+	track  bool
+}
+
+var (
+	mrCheckoutCfg mrCheckoutConfig
 )
 
 // listCmd represents the list command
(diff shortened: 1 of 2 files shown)


cmd/mrCheckout.go
```diff
@@ -36,18 +43,53 @@ var checkoutCmd = &cobra.Command{
...
+				}
+				err = git.RemoteAdd(mr.Author.Username, mrProject.SSHURLToRepo, ".")
+				if err != nil {
+					log.Fatal(err)
+				}
+			}
+			fetchToRef = fmt.Sprintf("refs/remotes/%s/%s", mr.Author.Username, mr.SourceBranch)
+		}
+
+		// https://docs.gitlab.com/ee/user/project/merge_requests/#checkout-merge-requests-locally
```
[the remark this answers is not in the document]
zaquestion: yikes, looks like I put that there :blush:

cmd/mrCheckout.go
```diff
@@ -36,18 +43,53 @@ var checkoutCmd = &cobra.Command{
...
 		err = gitf.Run()
 		if err != nil {
 			log.Fatal(err)
 		}
 
-		gitc := git.New("checkout", branch)
+		if mrCheckoutCfg.track {
+			// Create configured branch with tracking from fetchToRef
+			// git branch --flags <branchname> [<start-point>]
+			gitb := git.New("branch", "--track", mrCheckoutCfg.branch, fetchToRef)
```
zaquestion: It looks like the `--track` functionality can be used directly with `git checkout`. Is it worth combining the 2 commands?

internal/gitlab/gitlab.go
```diff
@@ -81,6 +81,18 @@ var (
 	localProjects map[string]*gitlab.Project = make(map[string]*gitlab.Project)
 )
 
+// GetProject looks up a Gitlab project by ID.
+func GetProject(projectID int) (*gitlab.Project, error) {
```
zaquestion: Maybe we should just update `FindProject` to accept an `interface{}` and `switch` for the type. `go-gitlab` will accept the interface form as well

cmd/mrCheckout.go
```diff
@@ -36,18 +43,53 @@ var checkoutCmd = &cobra.Command{
...
+			err = gitb.Run()
+			if err != nil {
+				log.Fatal(err)
+			}
+		}
+
+		// Check out branch
+		gitc := git.New("checkout", mrCheckoutCfg.branch)
```
zaquestion: Kind of a bad pattern on my part is looks like. These can really just be combined like
```
err := git.New("checkout", mrCheckoutCfg.branch).Run()
```

cmd/mrCheckout.go
adamryman: We would still need to check out this branch, as `git checkout --track -b` will track the new branch off of the previous branch. So if we use `git checkout --track` we need two `git checkout` commands.

I prefer this way personally.

cmd/mrCheckout.go
adamryman: Ooooooooooooooooooooooooooooo. I like that :+1:

internal/gitlab/gitlab.go
adamryman: After putting 15min into this. I request some help. The problem is resolving the whole `localProject` map.

How about I add a TODO? Maybe an issue as well? Or we could pair on it.

zaquestion: After chatting with you more (in person) this is fine as is.

zaquestion:
This just needs tests to merge.

zaquestion:
@adamryman you owe tests! :D

Issue #74, opened 2017-12-19
CI Pipeline interface


nkprince007:
**NOTICE: Description fully cannibalized by** @zaquestion
Still debating if such a thing should be in lab, but could always move it into its own tool and be called by lab.
Support
- Running jobs
- Viewing jobs status (live)
- Viewing environments
  - Include re-deploying jobs

Also add `lab ci lint` command
Ref: https://docs.gitlab.com/ce/api/lint.html

zaquestion:
What would the editor workflow look like?

zaquestion:
Also started working on a `ci` sub command. We could support this through `lab ci validate`

nkprince007:
Maybe we can't include editor workflow here, since the API doesn't show where the error is.

zaquestion:
![a_start](https://user-images.githubusercontent.com/2358914/34144360-123304de-e446-11e7-8faa-14826b46dbd6.png)
From: https://gitlab.com/zaquestion/test/pipelines/15273642

Using https://github.com/jroimartin/gocui

nkprince007:
That looks cool, can we maybe have some colours and arrows?

zaquestion:
Still static, but we're getting there!
![lab_ci](https://user-images.githubusercontent.com/2358914/34207592-032c7218-e540-11e7-9daf-d9d4c5989338.png)

From: https://gitlab.com/zaquestion/test/pipelines/15273642

zaquestion:
Put this up in an effort to support `lab ci lint`. https://github.com/xanzy/go-gitlab/pull/293

I've already written a working implementation with those changes. Once the API is exposed should be able to bring in the functionality in short order.

zaquestion:
Also relying on these changes for the pipeline interface: https://github.com/jroimartin/gocui/pull/130

zaquestion:
Have mostly replicated the current progress in https://github.com/rivo/tview which has been nicer to work with overall. 
![tview_ci_pipeline](https://user-images.githubusercontent.com/2358914/36411456-94fc1a26-15ca-11e8-90fe-2a18bbef6ad9.png)

- [ ] fix the rendering overall (notice the extra lines on the right most side)
- [ ] test live updating (implemented but not yet useful)
- [ ] screen doesn't clear when resizing (or between all updates?)
- [ ] Long term solution for drawing the lines (currently have tview hacked up to expose the screen to me)

zaquestion:
Latest look
[...]

PR #135, opened 2018-04-16
[#74] Add (beta) CI Pipeline interface: live view-only

zaquestion:
Updates every 5 seconds and displays how long jobs are running. With colors!


zaquestion:
Last thing here is identify a certain pesky unhandled panic that seems to happen when a pipeline exists but isn't "ready" usually retrying the command works, but by that time we've already borked your terminal

PR #166, opened 2018-06-14
(project_list) add command to list all projects

em-:
I've taken inspiration from `lab issue list` and wrote a `project` subcommand to list all the available projects.

It doesn't provide any useful filtering for the moment, but it was enough for my needs. :)


docs/lab_project_list.md
```diff
@@ -0,0 +1,23 @@
+## lab project list
```
zaquestion: The docs are actually Auto generated during release to keep them aligned to the latest released version. Can you remove this file?

cmd/project_list.go
```diff
@@ -0,0 +1,47 @@
...
+	lab "github.com/zaquestion/lab/internal/gitlab"
+)
+
+var projectListCmd = &cobra.Command{
+	Use:     "list [page]",
+	Aliases: []string{"ls"},
+	Short:   "List all projects",
+	Run: func(cmd *cobra.Command, args []string) {
+		page := 0
+		if len(args) == 1 {
```
zaquestion: Can probably use the `parseArgs` function defined in `cmd/root.go` to accept a search and/or page. Similar to how other commands accept a remote and/or page.

zaquestion: I beleive ListProjectOptions can accept such a search parameter, but I may be mistaken.

cmd/project_list.go
```diff
@@ -0,0 +1,47 @@
...
+	"log"
+	"strconv"
+
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+	"github.com/xanzy/go-gitlab"
+	lab "github.com/zaquestion/lab/internal/gitlab"
+)
+
+var projectListCmd = &cobra.Command{
```
zaquestion: Would like to see some sort of test, there are many examples in the `cmd` package.

cmd/project_list.go
zaquestion: This is optional, was just thinking out loud here.

zaquestion:
@em- I'm closing this out in favor of #173 I've picked up your changes (commit intact) and added search functionality + a test.

em-:
Thank you so much!

Issue #163, opened 2018-06-11
"lab create" = "hub create" when hub installed


d33tah:
When both `lab` and `hub` are installed, `lab create` actually runs `hub create`. This is super confusing - any chance you could make `lab XYZ` not try calling `hub XYZ`?

tydavis:
Maybe we should [stop integrating with hub][1].

[1]:https://github.com/zaquestion/lab#hub--lab--hublab

d33tah:
I would MUCH prefer it to also follow `hub`'s conventions. I don't mind `lab` having their own ones, but it would in my opinion help Gitlab adoption to have predictable CLI tools - even if you only just installed it.

zaquestion:
The approach I'm thinking right now is to stop having `lab` passthrough for this case. I agree its easy to run into and end up creating a github repo when you wanted a gitlab one. Since repo create is likely to be less frequent than other operations it seems fine to insist users call hub directly.

zaquestion:
@d33tah I've put up 2 PRs in an effort to address this issues. The first #167 removes the hub integration, the later #168 only prevents passthrough for the `hub create` command. Any preference from your side?

d33tah:
TBH, neither. I would prefer `lab create` to have same semantics as `hub create`, only for Gitlab. Thanks for writing the project, but you might have hard time pleasing me by introducing a new convention :"P

zaquestion:
@d33tah I'm hesitant, to allow `lab create` to work because it would break this passthrough to lab. We talked about this a bit when originally adding the feature https://github.com/zaquestion/lab/issues/113#issuecomment-370133273

Am interested in hearing your case, be aware lab hasn't "follow convention" for a some time. For a while the expectation was that hub would be moving in a similar direction (`cmd NOUN VERB`, see `hub pr`). lab does this to be inline with GitLabs documentation/API and aim to be discover-able. 

While lab started out with the goal of being true to `hub`, this is no longer the case. In order to take advantage of GitLab's offering and provide a clean API we've somewhat departed from that goal.

How does `lab create` serve you better than `lab project create`? Workflow is very important to us and drives virtually every decision made for lab.

d33tah:
@zaquestion It boils down to remembering things. Alternatively I would love to migrate to whatever is the alternative to`hub` that follows your syntax. Otherwise, I need to remember both and I don't use those tools often enough for it to become natural quickly, so it might be frustrating.

zaquestion:
[...]

Issue #149, opened 2018-05-04
Add `lab project browse` to open the project in the browser


zaquestion:
`cmd/issue_browse.go` and `cmd/mr_browse.go` can be used as reference here.

albfan:
I do almost everything from command line, but this hub feature is really appreciated for me.

I will take a look into those files, but: Will be feasible to detect origin is a gitlab repo and navigate to it directly?

zaquestion:
@albfan If I understand, yes, thats how its currently working for issues and merge requests. `lab project browse` would open your browse to origin by default or to the desired remote with `lab project browse <remote>`. This should be relatively straightforward to implement and there are some example tests IIRC.

albfan:
I reread my comment.

I mean if we can do

    lab browse

to behave as `hub browse` if it is a github repo or as `lab project browse` if it is a gitlab repo.

zaquestion:
`lab browse` since it is not recognized by lab will do a `hub browse` currently. It we want it to be able to do a `lab project browse` it's a bit extra work, but I'm not against it

zaquestion:
@albfan this is supported as off `0.12.0`, sorry for the delay!

PR #167, opened 2018-06-20
drop support for hub command

zaquestion:
not 100% about merging this, but issues like #163 and #149 show that integrating with hub complicates things and can produce unexpected results or confusion. Such as creating a repo on github, when you meant to create on on gitlab

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 <p align="center"><img src="https://user-images.githubusercontent.com/2358914/34196973-420d389a-e519-11e7-92e6-3a1486d6b280.png" align="center" height="350"></p>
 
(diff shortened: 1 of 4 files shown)


tydavis:
Given that this is a Gitlab command, it makes sense to have two different commands for different contexts. This might also open the door for more git plugin style use?

zaquestion:
I did go ahead with https://github.com/zaquestion/lab/pull/168 which should help prevent confusion. I'm fairly in agreement that git's plugin style would be nice, but I don't have a good way to support both atm and I strongly not wanting to do the extra typing when first looking for a tool like this.

In any case, I'm going to defer bringing these changes in for the time being, as personally I still get a lot of utility out of the passthrough and I think #168 strikes on a big part of the issue.

Issue #155, opened 2018-05-29
Load Config From Env


hadigoh:
Hello,
Is it possible to load config (host, token, email) from env var?
This is useful when using lab in a CI environment.
Or is there a way to give these values without reading from stdin?

zaquestion:
@hadigoh I thought we had support for this, but never tried it out myself, it looks like I need add a couple more lines in `main.go` to make it function. I'm hoping to do a release soon and get a bunch of these relatively easy feature requests in, will include this.
```
  viper.SetEnvPrefix("LAB")
  viper.SetEnvKeyReplacer(strings.NewReplacer(".", "_"))
```

zaquestion:
@hadigoh As a work around you can always write the lab config file and load config for that. Lab will use the config file in the current directory if present or `~/.config/lab.hcl`

zaquestion:
This is now supported in `0.12.0` with `LAB_` prefix and underscores in place of `.` for nested keys. e.g. `LAB_CORE_HOST`

PR #186, opened 2018-07-21
load config from file only if env is not present

hadigoh:
Hello @zaquestion 
Continuing #155, even after setting the required env (LAB_CORE_HOST, LAB_CORE_USER, LAB_CORE_TOKEN), lab still asks to create the config file if it's not present.

So this pull request handles if no config file is present but the config is in env, then it wont asks user to input the config from stdin, but return the values from env instead.

And `viper.AllSettings()` seems to not return config from env, hence this pr checks for env and return the config asap


zaquestion:
Thanks!

Issue #187, opened 2018-07-23
Trigger pipeline from lab


temp:
Would be nice to be able to trigger a pipeline from lab.

Something like:
```
lab ci run my/project master
```

PR #188, opened 2018-07-24
Introducing CI run command to create a pipeline

dimitrovvlado:
Introducing command to start/run a pipeline. The current implementation uses the current branch/ref and assumes that the remote branch has the same name - this can be changed with additional argument to the command. Also not sure if `run` is the best name for the command, `create` is the verb chosen by gitlab.

This pull request addresses https://github.com/zaquestion/lab/issues/187

First time contributing in this project, so not sure if master is the correct target branch :)

### DIFF
--- a/cmd/ci_run.go
+++ b/cmd/ci_run.go
@@ -1,0 +1,39 @@
+package cmd
+
+import (
+	"log"
+
+	"github.com/spf13/cobra"
+	gitlab "github.com/xanzy/go-gitlab"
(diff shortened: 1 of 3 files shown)


zaquestion:
@dimitrovvlado Thanks this looks great! Heads up, I may update the output to print the URL over just the pipeline_id before releasing, but this is great.

zaquestion:
Against master is perfect. During release we tag a Sha, diff the last release, and generate documentation. It's safe for the code in master to be ahead of the latest release.

PR #193, opened 2018-08-01
Touchup makefile

alerque:
Here are a couple things I fixed up while messing around building this on my own systems and [packaging it for ArchLinux](https://aur.archlinux.org/packages/lab). Golang projects are actually a pain in the butt to package because they are so adamant about doing everything themselves (downloading sources, verifying checksums, building form only verified remote paths, etc.) that it's hard to get a package manager wrapped around them. The more bits of the process have standard `make` or similar wrappers the better, and splitting these dependencies so they can be run at the corresponding time in the package process would be better than the monolithic approach.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,14 @@
-install:
+VERSION ?= $(shell git rev-parse --short=10 HEAD)
+GOURL ?= github.com/zaquestion/lab
+
+deps:
 	dep ensure
-	go install -ldflags "-X \"main.version=$$(git rev-parse --short=10 HEAD)\"" github.com/zaquestion/lab
+
+build: deps
+	go build -ldflags "-X \"main.version=$(VERSION)\"" $(GOURL)
+
+install: build
+	go install $(GOURL)
 
 test:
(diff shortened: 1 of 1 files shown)


alerque:
Eww gross. Let me rebase to fix that merge conflict in a less messy way that Github's tool.

zaquestion:
@alerque thanks for the touch up! As a side note I'd love to automate the packaging for the arch package. I've been vaguely aware someone has been maintaining the arch package since v0.5, let me know if I can ease that process!

alerque:
I am now one of the maintainers for both the `lab` and `lab-git` packages and will be cleaning them up. Splitting the build steps as in this PR so that each step can be run (and verified) by the package management system in sequence will make it easier. If there are more things that can be done to simplify it I'll be sure to suggest or submit them here too. The biggest hassle is that the `go` builder wants to download sources on its own while the package system really wants to use it's own copies of sources that it has verified match the checksums it keeps and prefers the build to happen in a sandbox.

PR #195, opened 2018-08-02
(listing) abstract pagination from user

zaquestion:
Users can either use `-a` to fetch all results or `-n` to specify a number of results


zaquestion:
There's a bug in this, doesn't perfectly stop on the correct number of results. Need to adjust page side for last batch

PR #210, opened 2018-08-10
fix remove source branch and allow collaboration mr create options

doits:
refactored only one place, so the underscores should be hyphens instead


doits:
yeah, just tried it out in 0.13 and wondered why it didn't work :-/

zaquestion:
@doits released this in `0.13.1`, that KS again for the quick fix

doits:
no problem: just fixing the tools I'm using :-) (and my own error)

PR #222, opened 2018-09-25
Allow filtering by target branch

marionebl:
This introduces a feature for `mr list`, where users can filter the list of printed merge requests by target branch.

### DIFF
--- a/cmd/mr_list_test.go
+++ b/cmd/mr_list_test.go
@@ -101,7 +101,7 @@ func Test_mrFilterByTargetBranch(t *testing.T) {
 	}
 
 	mrs := strings.Split(string(b), "\n")
-	require.Equal(t, "PASS", mrs[0])
+	require.Equal(t, 3, len(mrs))
 }
 
 func Test_mrListByTargetBranch(t *testing.T) {
@@ -116,5 +116,5 @@ func Test_mrListByTargetBranch(t *testing.T) {
 	}
 
 	mrs := strings.Split(string(b), "\n")
-	require.Equal(t, "#107 WIP: Resolve \"issue title\"", mrs[0])
+	require.Equal(t, "#1 Test MR for lab list", mrs[0])
 }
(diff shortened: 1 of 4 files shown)


marionebl:
A hint about differences between local and remote testing would be greatly appreciated :)

cmd/mr_list_test.go
```diff
@@ -88,3 +88,33 @@ func Test_mrListFivePerPage(t *testing.T) {
...
+	cmd := exec.Command("../lab_bin", "mr", "list", "-t", "non-existing")
+	cmd.Dir = repo
+
+	b, err := cmd.CombinedOutput()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	mrs := strings.Split(string(b), "\n")
+	require.Equal(t, "PASS", mrs[0])
```
zaquestion: "PASS" will always be present in the output (assuming non-zero exec which will show as an err to the exec call). Its better to assert that the length only has the test prints in it. I think there are 2? lines. so something like `require.Equal(t, 2, len(mrs))

zaquestion:
@marionebl thanks for contributing. It looks like the GitLab API supports filtering by branch name https://docs.gitlab.com/ee/api/merge_requests.html#list-merge-requests

You should just be able to specify `TargetBranch` in the just above where the MRs are listed `gitlab.ListProjectMergeRequestsOptions`

Testing wise you should just need an ssh key attached to a gitlab account for `make test` to work and you can use `make test run=<regex>` (same as you'd use with go test -run=). The data its looking at lives in https://gitlab.com/zaquestion/test/merge_requests

marionebl:
> @marionebl thanks for contributing. It looks like the GitLab API supports filtering by branch name https://docs.gitlab.com/ee/api/merge_requests.html#list-merge-requests

> You should just be able to specify TargetBranch in the just above where the MRs are listed gitlab.ListProjectMergeRequestsOptions

Being new to Go as an ecosystem I originally wanted to avoid the go-gitlab update required to do this, gave it a try now. 

It appears to break some other functionality, though: https://travis-ci.org/zaquestion/lab/builds/433313336#L567

I adapted the test case as suggested.

marionebl:
Fixed the issue with `ApprovalsBeforeMerge` by defaulting it to `0`

PR #224, opened 2018-10-18
Add liberapay donation link

MartinDelille:
I made a donation on Liberapay. Why not adding a badge on the README.md file?

I also add carriage return between the badges. It doesn't changed the rendered markdown but it make the source clearer.


zaquestion:
@MartinDelille Wow thanks! For some reason the newlines seems to cause the badges to end up on the next line. In the few minutes I fiddled with I couldn't figure out how to keep the nice formatting and get the badges on the same line as the titles. Feel free to send another update if you figure it out!

Issue #76, opened 2017-12-19
Support  edit functionality on mrs, issues, and snippets


zaquestion:
Wondering if this should be a `edit` sub command or a flag `--amend` (like git)

Want to support setting/updating properties without having to edit the message. The flag approach would make sense to use `--no-edit`. Not sure the best choice for the sub command route...

That said, leaning towards the sub command to be consistent with other commands, and optionally support `--amend` as the flag off the parent command.

cuthix:
I very often just need change text or tick/untick part of issue, so would be nice to have command/flag to edit issue text :)

claytonrcarter:
@zaquestion I'm tinkering with this and would like some feedback on the behavior and and flag naming. I'm aiming for parity/familiarity with the various `lab ... create` commands, i.e. title/description/labels/assignees. Here's the help text so far:

```
$ ./lab issue edit -h
Edit or update an issue

Usage:
  lab issue edit [remote] <id> [flags]

Aliases:
  edit, update

Flags:
  -a, --assign strings     Add an assignee by username
      --edit               Open the issue in an editor (default if no other flags given)
  -h, --help               help for edit
  -l, --label strings      Add the given label(s) to the issue
  -m, --message strings    Use the given <msg>; multiple -m are concatenated as separate paragraphs
  -t, --title string       Set the issue title
  -A, --unassign strings   Remove an assigne by username
  -L, --unlabel strings    Remove the given label(s) from the issue
```

### Specific Questions

1. I wanted to add a flag to just change the title, hence `--title`, but this seems to beg for `--description` flags, too. How would something like that interact with the `--message` flags (which are standard on `lab issue create`, `lab issue note`, `lab mr create`, etc). Adding `--description` would be more declarative, which might be appropriate for an `edit` command, but it could be confusing if `--message` is left in place. On the other hand, if `--message` is removed, the syntax would break from the `create` commands, which doesn't sound great either.
2. Do the capital shorthands for unassigned and unlabel sound OK?

### Usage Examples

Edit issue in editor: `lab issue edit <id>`  
Only update issue title:  `lab issue edit <id> -t "new title"` or: `lab issue edit <id> -m "new title"`  
Only add label to issue:  `lab issue edit <id> -l newlabel`  
Add label to issue and edit title/description in editor:  `lab issue edit <id> -l newlabel --edit`

zaquestion:
[...]

PR #245, opened 2018-11-14
Issue edit

claytonrcarter:
See discussion in #76. I skipped the `title` and `edit` flags, and removed the shorthand for `unlabel` and `unassign`. 

This is fully functional, but before it goes further we should confirm the naming of the flags, and specifically the assign & unassign flags. They pair nicely, of course, and I like how they feel in use but they're inconsistent with other commands: `lab issue create` uses `--assignees` (even though it only takes one) and `lab mr create` uses `--assignee`. If I change this to use `assignee`, then `unassign` feels out of place, but changing `unassign` to something like `remove-assignee` feels overly verbose.  Feedback or opinions would be helpful.

```
$ ./lab issue edit -h
Edit or update an issue

Usage:
  lab issue edit [remote] <id> [flags]

Aliases:
  edit, update

Examples:
lab issue edit <id>                                # update issue via $EDITOR
lab issue update <id>                              # same as above
lab issue edit <id> -m "new title"                 # update title
lab issue edit <id> -m "new title" -m "new desc"   # update title & description
lab issue edit <id> -l newlabel --unlabel oldlabel # relabel issue

Flags:
  -a, --assign strings     Add an assignee by username
  -h, --help               help for edit
  -l, --label strings      Add the given label(s) to the issue
  -m, --message strings    Use the given <msg>; multiple -m are concatenated as separate paragraphs
      --unassign strings   Remove an assigne by username
      --unlabel strings    Remove the given label(s) from the issue
```


cmd/issue_edit_test.go
```diff
@@ -0,0 +1,234 @@
...
+	return s[len(s)-1]
+}
+
+func issueEditCmdTest_showIssue(dir string, issueNum string) string {
+	cmd := exec.Command("../lab_bin", "issue", "show", "lab-testing", issueNum)
+	cmd.Dir = dir
+
+	b, err := cmd.CombinedOutput()
+	if err != nil {
+		log.Fatal(err)
```
zaquestion: consider passing the `*testing.T` to these helper function so you can do a `t.Fatal(err)` here instead of the `log.Fatal(err)`

claytonrcarter:
I just updated this use to use `--assignee` instead of `--assign` to be more consistent with `mr create`. Anything else I can do to help move this along?

zaquestion:
@claytonrcarter Sorry I'm just looking at this again since last time. I think `--assign` make more sense here, because issues support multiple assignees while merge requests only support 1. When merge requests get updated to support multiple (I think still pending?) I figured we'd update merge requests to use `--assign` as well.

Do you think it makes sense to change back with that in mind? Read to merge this when I hear back.

zaquestion:
@claytonrcarter forgot I added you as a collaborator. Approved the PR, feel free to merge at your discretion.

claytonrcarter:
Makes sense! I reverted the last commit and will merge later today. Thanks!

Issue #40, opened 2017-11-09
Add shell completion scripts


zaquestion:
Cobra has some docs on generating autocompletion scripts for bash: https://github.com/spf13/cobra/blob/master/bash_completions.md

zaquestion:
Using the cobra generated bash completion is ended up being rather difficult to use in conjunction with git's bash completion. Moreover, bash completion is clearly well beyond my meager mind. I did put up some work at making a completion script by hand in https://github.com/zaquestion/lab/commit/e1afa8be814976ccc1b96cc01568ceed34dc887a which supports completing the commands but not the arguments.

sumnerevans:
It would be nice to create zsh completions as well.

claytonrcarter:
I'm working on completions for fish, which is actually fairly straightforward, but it's also all manual and ... a little tedious. I see that there is [a pending PR to enable fish completions in cobra](https://github.com/spf13/cobra/pull/754) but I'm still too fresh to go and cobra to understand what the docs are saying. I'll post my completions when I get them settled.

zaquestion:
Can't make any promises, but Ill try and run the code in that PR and generate the fish completion for you. Hopefully it'll at least be a good jumping off point, but if it's generating good completions then we might be able to poke some folks for a merge

zaquestion:
Something to keep in mind here is that lab will no longer be a git wrapper in 1.0 -- instead we'll aim to use "git extensions" (or w/e their called) with aliases in git to achieve the same result. Hopefully this should simplify the completetion problem dramatically. In any cases it's likely not worth the headache to get completions working in conjunction with gits completions. I'd suggest focusing on just completing labs commands

zaquestion:
@claytoncarter an example of using cobras completion generators can be seen in https://github.com/zaquestion/lab/pull/205/files

claytonrcarter:
Thanks, and that makes sense, especially if completions are a pain to get right in bash. FWIW, fish makes it trivial to "inherit" completions from other commands: `complete -f -c lab -w git` tells fish to "wrap" `git` when completing the `lab` command.

zaquestion:
:eyes:

rsteube:
Got some basic zsh completion working: https://github.com/rsteube/lab-completion

zaquestion:
@rsteube Thats exciting, if the completions work I'm open to merging the changes into lab (even if that means using a fork of cobra). I'm sure the community would really appreciate completion scripts.

[...]

PR #252, opened 2018-11-27
Added bash and zsh autocompletion

rsteube:
Added `bash` and `zsh` autocompletion generation using cobra. This is based on the currently ongoing PR by babysnakes which i forked to support zsh custom completion for positional arguments.
When babysnakes PR finally lands in cobra, the changes shouldn't differ too much.

The `bash` version simply completes the commands while the `zsh` one has some additional argument completion (like menu completion of issue numbers with their title).

- ~~added Gopkg files (dep)~~
- ~~fixated 2 dependencies to an older commit in lock file (build problem)~~
- changed cobra source to my personal fork
- added `lab completion {bash,zsh}` command
- added some initial positional argument completions using lab and git invocations

The `remote` argument of some commands is currently skipped. I started working on that, but this might be a good first iteration since most of it seems to work so far.

see #40 
changes made in my cobra fork can be seen here: https://github.com/babysnakes/cobra/compare/zsh-completion...rsteube:zsh-completion-custom


rsteube:
Seems the build fails because it doesn't use `dep`: https://travis-ci.org/zaquestion/lab/builds/460512521#L498

zaquestion:
Yeah we switch over to go modules recently. Likely you just need to upgrade to go1.11 and do something like `GO111MODULE=on go build .` and the go.mod file should get updated. Thanks for putting this together, I'll try to pick it up this weekend if we're still facing issues.

rsteube:
Ah, didn't know about modules yet. I was looking for something like that but it seems i was blind and didn't expect the files to be called `*.mod` and `*.sum`. Added a `replace` instruction to the `go.mod`.

PR #254, opened 2018-11-29
add positional completion to issue edit

claytonrcarter:
#245 landed just after #252, so it missed the completion party. I think that this adds the required code to enable positional completion of `[remote]` and `<id>`. @rsteube can you take a peek to make sure I'm doing it right? I ~~copi~~...based this off of the completions in `issue note`.

Thanks!


rsteube:
Yes that's correct. Tried it out and it works.
(fyi: https://github.com/zsh-users/zsh-completions/blob/master/zsh-completions-howto.org#utility-functions)

Issue #250, opened 2018-11-27
docs: Update help text for `lab project create`


claytonrcarter:
I would like to update the help text of `lab project create`. Here is the current output:

```
$ ./lab project create -h
If no path or name is provided the name of the git repo working directory

path refers the path on gitlab including the full group/namespace/project. If no path or name is provided and the directory is a git repo the name of the current working directory will be used.

Usage:
  lab project create [path] [flags]

Flags:
  -d, --description string   description to use for the new project
  -h, --help                 help for create
  -n, --name string          name to use for the new project
  -p, --private              Make project private; visible only to project members (default: internal)
      --public               Make project public; can be accessed without any authentication (default: internal)
```

The first sentence is incomplete and the second sentence is incorrect: as far as I can tell, the path currently cannot include the group/namespace. (I get an API error response when I include the namespace.) Finally, the wording of `--private` and `--public` seems clumsy and they seem to imply that those flags take values like `--public internal` or `--public ???`.

My suggested revisions:

```
$ ./lab project create -h
Create a new project on GitLab in your user namespace.

path refers to the path on GitLab not including the group/namespace. If no path
or name is provided and the current directory is a git repo, the name of the
current working directory will be used.

By default, projects are created with 'internal' visibilty (visible to any
authenticated user). Projects can also be made 'public' or 'private' via flags.

Usage:
  lab project create [path] [flags]

Examples:
# this command...                          # creates this project
lab project create                         # user/<curr dir> named <curr dir>
                                           # (above only works w/i git repo)
lab project create myproject               # user/myproject named myproject
lab project create myproject -n "new proj" # user/myproject named "new proj"
lab project create -n "new proj"           # user/new-proj named "new proj"

Flags:
  -d, --description string   description of the new project
  -h, --help                 help for create
      --internal             Make project internal: visible to any logged in user (default)
  -n, --name string          name of the new project
  -p, --private              Make project private: visible only to project members
      --public               Make project public: visible without any authentication
```

Changes:

1. rewrote the main help block, adding mention of the default visibility. I don't like that it's wordier, but thought it was worth discussing.
2. added Examples to clarify how `path` and `name` relate
3. removed "default" statements from `--private` and `--public`
4. This ones a little weird: I added an `--internal` flag, with description and note that it's the default. It's a noop and doesn't actually do anything in the code; it's just an idea and only there to document the default. My thinking is that we use either this noop flag *or* the added visibility text above, but not both.

zaquestion:
[...]

PR #255, opened 2018-11-29
project create: update help text

claytonrcarter:
Based on #250, I removed the verbiage from the help text about visibility and left in the `--internal` flag to do the explaining for us. While I was at it, I made `--internal` an actual flag b/c I was worried that someone might (mistakenly) try to use `--internal --public`. In the case of multiple visibility flags, the command with use the "most private" of those provided.

### DIFF
--- a/cmd/project_create.go
+++ b/cmd/project_create.go
@@ -11,13 +11,24 @@ import (
 	lab "github.com/zaquestion/lab/internal/gitlab"
 )
 
+// private and public are defined in snippet_create.go
+var internal bool
+
 // projectCreateCmd represents the create command
 var projectCreateCmd = &cobra.Command{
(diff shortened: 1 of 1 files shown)


cmd/project_create.go
```diff
@@ -29,10 +40,17 @@ path refers the path on gitlab including the full group/namespace/project. If no
...
+		// set the default visibility
 		visibility := gitlab.InternalVisibility
+
+		// now override the visibility if the user passed in relevant flags. if
+		// the user passes multiple flags, this will use the "most private"
+		// option given, ignoring the rest
 		switch {
 		case private:
 			visibility = gitlab.PrivateVisibility
+		case internal:
```
zaquestion: good catch!

PR #263, opened 2018-12-13
install.sh script fixes

### LINKED ISSUE #262: Install to user space
The install script installs into a system bin directory. Would it be acceptable to propose a PR to allow some user configuration here?

The use case is basically that there are users that might not have `sudo` privileges on their machine to install into `/usr/local/bin`, while it is perfectly acceptable to install to `~/.local/bin` (via CLI argument to `install.sh`).

1ace:
Started looking at this script to fix #262, and fixed a few other things while there :)

installing to a custom location is now done by setting the environment variable `PREFIX` before executing the script; eg.:
```bash
PREFIX=/usr install.sh
```
or
```bash
PREFIX=~/.local install.sh
```

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -55,4 +55,5 @@ latest="$(curl -sL 'https://api.github.com/repos/zaquestion/lab/releases/latest'
 
 curl -sL "https://github.com/zaquestion/lab/releases/download/v${latest}/lab_${latest}_${os}_${machine}.tar.gz" | tar -C /tmp/ -xzf -
 cp /tmp/lab $BINDIR/lab
+chmod +x $BINDIR/lab
 echo "Successfully installed lab into $BINDIR/"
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -1,16 +1,35 @@
...
-set -e
+set -eu -o pipefail
 
-if [[ ! -z $DEBUG ]]; then
+if [[ ! -z ${DEBUG-} ]]; then
     set -x
 fi
 
-if [[ $EUID != 0 ]]; then
+: ${PREFIX:=/usr/local}
```
claytonrcarter: Sorry, but my bash is a little weak. Would you mind explaining what this line is doing?

1ace: Sure:
- the `:` makes the rest of the line evaluate but discards the output (we don't need the variable value there)
- the `${VAR:=value}` construct is a shell variable expansion which will check if `$VAR` is set (and not empty), and if it isn't it will assign it the `value` given

Issue #261, opened 2018-12-06
ci view: short or brief output?


claytonrcarter:
Not sure if it's just my use case or if it would be useful for others, but I often just want to get a quick gist of what CI is currently up to. `ci view` provides this of course, but if I don't want to trace or interact with jobs, it feels like overkill, plus I have to `q` out of it. What I would love is something like `lab ci status` or maybe `lab ci view --brief` that just lists the job names and their current status. For example (where I only have one job named "test"):
```
$ git push
...
$ lab ci status
test: pending
$ lab ci status
test: running
$ lab ci status
test: success
```

zaquestion:
:+1: to a `lab ci status` I've wanted one for a while and something like the output you're describing is appealing. It seems like the biggest challenge dealing with the fact that pipelines have multiple jobs. So if you're proposing just doing a textual semi-ordered list that seems reasonable
```
$ lab ci status
build: success
test:  running
deploy: pending
```
Actually sounds pretty rad!

nkprince007:
I totally second this approach. (adding a few emojis makes it even cute though)

```
$ lab ci status
✅ build: success
🏃🏻‍♂️ test: running
🕐 deploy: pending
❌ e2e: failed
⏸ docker-build: paused
```

MartinDelille:
I posted a proposition here: https://github.com/zaquestion/lab/issues/259#issuecomment-447343828

claytonrcarter:
Closed (a while ago) by #273

Issue #240, opened 2018-11-07
lab ci wait


temp:
I don't know wether or not this is possible, but it would be nice to have a ci command that simpy waits for the current pipeline to finish and then exits with status code 0 for success, and 1 for failure.
This would allow scripting of several pipelines that need to be run in a specific order. I know you can trigger builds inside a pipeline, but we need it the other way - like a listener.

This would allow something like this:
```
cd /whatever/docker-base
lab ci create
lab ci wait
cd /whatever/docker-project-that-uses-docker-base
# or maybe
lab ci create --wait
cd /whatever/another-docker-project-that-uses-docker-base
# or maybe
lab ci create
lab ci trace --exit-on-finish
...
```

zaquestion:
I'm pretty sure we can pull this off! Thanks for including your usecase as well, that goes a long way in integrating your idea into lab as a whole and provide a tangible workflow to justify implementation.

That said, have you looked at multi-project pipelines in GitLab? This may fit your usecase more elegantly that the feature would?? https://docs.gitlab.com/ee/ci/multi_project_pipelines.html -- Let me know!

temp:
Yes, I looked at multi project pipelines, but they don't support our use case.
We don't need a trigger from upstream, but a listen from downstream.
Like in my example a base docker image.
Now for example 10 projects use this base docker image, and every time the base docker image is built, I want the 10 project docker images to be rebuild. But I don't want to add trigger code for each project in the base docker repository...

zaquestion:
@temp `lab ci status --wait` was added in `0.15.0` please let me know if this works for your purpose and reopen if needed.

temp:
Looking good!
Maybe the spam-factor can be reduced by waiting a second between status-check while --wait'ing? Or add something like --quiet?

temp:
Ah, some weirdness remains ;-)
I have a status --wait running, and it's still printing running messages, but a status on a second console reveals:

```
± lab ci status
Stage:   Name              - Status
...
Pipeline Status: failed
```

Maybe the wait is already done, but he's still busy printing?

temp:
Wow, it kept running, and Mac OS just complained about no app memory left... 
Here's what it said:
![bildschirmfoto 2019-02-11 um 11 34 24](https://user-images.githubusercontent.com/216128/52557906-21dcd680-2df1-11e9-8e76-0f59be9a9550.png)
:-D

zaquestion:
Oh my!

[...]

PR #273, opened 2018-12-28
(ci status) support a short text based pipeline/job status

zaquestion:
To address #261 & through the use of `--wait` hopefully #240


zaquestion:
@claytonrcarter @nkprince007 @MartinDelille  need to actually add some tests, but what are your thoughts on this?

Example output
```
unittest: unittest - success
build: docker - success
sandbox: deploy_s-s0 - success
sandbox: deploy_s-s1 - manual

Pipeline Status: success
```

nkprince007:
> @claytonrcarter @nkprince007 @MartinDelille need to actually add some tests, but what are your thoughts on this?
> 
> Example output
> 
> ```
> unittest: unittest - success
> build: docker - success
> sandbox: deploy_s-s0 - success
> sandbox: deploy_s-s1 - manual
> 
> Pipeline Status: success
> ```

Looks good, but I'd prefer having them aligned properly.

Like so:
```
unittest: unittest - success
build:    docker - success
sandbox:  deploy_s-s0 - success
sandbox:  deploy_s-s1 - manual

Pipeline Status: success
```

Reference: https://golang.org/pkg/text/tabwriter/

claytonrcarter:
I agree with @nkprince007 about aligning the job names. That said, I don't think such a small tweak should hold this up. It's simple, clean and works! A followup PR would to change the alignment would be easy enough.

Do you have any ideas about how to test it?

zaquestion:
Yeah easy to test just need to actually do it. Should be able to get the in this weekend.

zaquestion:
@nkprince007 @claytonrcarter I've added some tests and aligned the text with `text/tabwriter`

claytonrcarter:
Love it!

PR #275, opened 2018-12-31
Make all tests pass on Windows.

matthewdale:
- Remove any explicit references to executables or paths that are not available on a standard Windows installation
- Run all tests in copies of the testdata dir; do not modify the main testdata dir
- Clean paths in tests to make Unix-style and Windows-style paths comparable
- Move some logic from the Makefile into the Go test driver, allowing tests to be run without Make
- Remove the Test_snippetCreate_Global_Editor test, which seems unlikely to work on Windows and of dubious value
- Update CONTRIBUTING.md to reflect reduced reliance on make

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,4 +13,7 @@ _test
 
 coverage*
 vendor
+
+# The testdata-* copies created by the unit tests must be ignored or the tests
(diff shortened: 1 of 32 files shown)


matthewdale:
Resolves #276

PR #283, opened 2019-01-29
WIP: Add `--patch` option to `mr show`

### LINKED ISSUE #277: `git lab mr show` should be able to show the actual MR
Right now `git lab mr show` only shows a summary with the MR title and description, and a couple meta-data, but not the content of the MR.

It would make it much more useful if there was an additional flag (say `-p, --patch` for instance) that would fetch the branch and show it the way `git show ..$remote/$branch` does.  
(Bonus point for supporting `--reverse` or even making it the default, because the "anti-chronological order by default" of `git show` annoys me.)

1ace:
Fixes #277

I also need to write tests for all the bits added; will do if/when you guys confirm that this is the right approach.

Also, an open question: do we want to mimic the behavior of `git show` with the `--reverse` flag, or can we just make it always pass `--reverse` so that the commits always show up in chronological order?
I'd prefer the latter, but I understand if we want the former for consistency.

### DIFF
--- a/cmd/mr_show.go
+++ b/cmd/mr_show.go
@@ -7,7 +7,13 @@ import (
 
 	"github.com/spf13/cobra"
 	gitlab "github.com/xanzy/go-gitlab"
+	"github.com/zaquestion/lab/internal/git"
 	lab "github.com/zaquestion/lab/internal/gitlab"
+)
+
+var (
+	mrShowPatch        bool
+	mrShowPatchReverse bool
 )
 
 var mrShowCmd = &cobra.Command{
@@ -27,7 +33,16 @@ var mrShowCmd = &cobra.Command{
 			log.Fatal(err)
 		}
 
-		printMR(mr, rn)
+		if mrShowPatch {
+			remote := args[0]
+			err := git.Fetch(remote, mr.SHA)
+			if err != nil {
+				log.Fatal(err)
+			}
+			git.Show(remote+"/"+mr.TargetBranch, mr.SHA, mrShowPatchReverse)
+		} else {
+			printMR(mr, rn)
+		}
 	},
 }
 
@@ -70,6 +85,8 @@ WebURL: %s
 }
 
 func init() {
+	mrShowCmd.Flags().BoolVarP(&mrShowPatch, "patch", "p", false, "Show MR patches")
+	mrShowCmd.Flags().BoolVarP(&mrShowPatchReverse, "reverse", "", false, "Reverse order when showing MR patches (chronological instead of anti-chronological)")
 	mrShowCmd.MarkZshCompPositionalArgumentCustom(1, "__lab_completion_remote")
 	mrShowCmd.MarkZshCompPositionalArgumentCustom(2, "__lab_completion_merge_request $words[2]")
 	mrCmd.AddCommand(mrShowCmd)
--- a/internal/git/git.go
+++ b/internal/git/git.go
@@ -110,6 +110,29 @@ func Log(sha1, sha2 string) (string, error) {
 	return string(outputs), nil
 }
 
+// Fetch a commit from a given remote
+func Fetch(remote, commit string) error {
+	gitcmd := []string{"fetch", remote, commit}
+	cmd := New(gitcmd...)
+	cmd.Stdout = nil
+	cmd.Stderr = nil
+	err := cmd.Run()
+	if err != nil {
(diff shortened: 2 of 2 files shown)


zaquestion:
@1ace This direction seems solid and I'm favorable to matching git ordering. lab tries to provide a got like syntax to most commands. 

Long term I'm looking to have mr diffing functionality in a `lab mr review` command, but the functionality you have her seems appropriate and fits well in a show command

Adding functionality to auto detect the MR from the branch would be appreciated as well. I believe there's an implementation lying around mr browse

zaquestion:
@prarit has been working on the `mr show` command lately. Any thoughts on these changes?

prarit:
I like this idea, however, two things come to mind.

1) As noted above it would be good to have test code written for this new functionality.  That should be easy to do with the test repo.
2) I wish there was something we could do to automatically find the remote as having to specify that seems strange on the command line. If I do a 'lab mr show 22' I don't have to specify the remote but I do for 'lab mr show 22 origin --patch'.  I wonder if there is some slick way we could automate the remote?
3) Should --patch do a 'git log -p'/'git show' or should it actually do a git-format-patch?  I feel like git-format-patch would be more useful for reviewers.  OTOH, reviewers could do 'lab mr checkout 22' (again notice no remote was specified) followed by a 'git-format-patch'.

Thoughts?  I like where this is headed but I think a little extra work would make it perfect.

cmd/mr_show.go
```diff
@@ -27,7 +33,16 @@ var mrShowCmd = &cobra.Command{
 			log.Fatal(err)
 		}
 
-		printMR(mr, rn)
+		if mrShowPatch {
+			remote := args[0]
```
prarit, line 37:
```suggestion
			var remote string

                        if len(args) == 1 {
                                remote = findLocalRemote(mr.TargetProjectID)
                        } else if len(args) == 2 {
                                remote = args[0]
                        } else {
                                log.Fatal("Too many arguments.")
                        }

```

cmd/mr_show.go
```diff
@@ -27,7 +33,16 @@ var mrShowCmd = &cobra.Command{
...
+			err := git.Fetch(remote, mr.SHA)
+			if err != nil {
+				log.Fatal(err)
+			}
+			git.Show(remote+"/"+mr.TargetBranch, mr.SHA, mrShowPatchReverse)
+		} else {
+			printMR(mr, rn)
+		}
 	},
 }
```
prarit, line 48:
```suggestion
func findLocalRemote(ProjectID int) (string) {
        var remote string

        project, err := lab.GetProject(ProjectID)
        if err != nil {
                log.Fatal(err)
        }
        remotes_str, err := git.GetLocalRemotes()
        if err != nil {
                log.Fatal(err)
        }
        remotes := strings.Split(remotes_str, "\n")

        // find the matching local remote for this project
        for r  := range remotes {
                // The fetch and push entries can be different for a remote.
                // Only the fetch entry is useful.
                if strings.Contains(remotes[r], project.SSHURLToRepo + " (fetch)") {
                        found := strings.Split(remotes[r], "\t")
                        remote = found[0]
                        break
                }
        }

        if remote == "" {
                log.Fatal("remote for", project.SSHURLToRepo, "not found in local remotes")
        }
        return remote
}

```

internal/git/git.go
```diff
@@ -110,6 +110,29 @@ func Log(sha1, sha2 string) (string, error) {
...
+// Show all the commits between 2 git commits
+func Show(commit1, commit2 string, reverse bool) {
+	gitcmd := []string{"show"}
+	if reverse {
+		gitcmd = append(gitcmd, "--reverse")
+	}
+	gitcmd = append(gitcmd, fmt.Sprintf("%s..%s", commit1, commit2))
+	New(gitcmd...).Run()
+}
+
```
prarit, line 135:
```suggestion

// GetLocalRemotes returns a string of local remote names and URLs
func GetLocalRemotes() (string, error) {
        cmd := New("remote", "-v")
        cmd.Stdout = nil
        remotes, err := cmd.Output()
        if err != nil {
                return "", err
        }

        return string(remotes), nil
}

```

prarit: These changes look at the MR's project URL and maps that to name of the fetch repo in 'git remote -v'.  That way 

lab mr show 9 --patch 

and

lab mr show origin 9 --patch

both work.

prarit:
I have incorporated these changes into a new pull request 413.

PR #293, opened 2019-02-12
(ci trace) copy to the writer, dont assume stdout, oops

zaquestion:


zaquestion:
tests passed on PR, not sure why status isn't reporting back, but not going to wait on it.

PR #298, opened 2019-02-26
Fix spelling error

juliangilbey:
Just a tiny typo


juliangilbey:
I forgot to say: this was spotted by Felix Lechner <felix.lechner@lease-up.com>

PR #299, opened 2019-02-26
tview patch: replace ANSII by ANSI

juliangilbey:
The tview package has replaced ANSII* by ANSI* everywhere; this patch makes lab build with the current tview package.

This patch is due to Felix Lechner <felix.lechner@lease-up.com>


juliangilbey:
Oh, actually, this patch would only work if the dependency is updated to a newer version of tview.  Hmmm....

zaquestion:
We can update the tview dep. There are decent tests in place for asserting the screens are rendered correctly.

zaquestion:
Closing in favor of https://github.com/zaquestion/lab/pull/322 which builds on these changes and updates tview

PR #312, opened 2019-06-04
Fix crash when opening issue in browser

letientai299:
Lab crashes due to loading a non existed config file for the host url. Since other commands work just fine without the config file, we can use similar approach for `issue browse` command, which is detect project url from git host and build up the url.

### DIFF
--- a/cmd/issue_browse.go
+++ b/cmd/issue_browse.go
@@ -2,7 +2,6 @@ package cmd
 
 import (
 	"log"
-	"path"
 	"strconv"
 
 	"github.com/spf13/cobra"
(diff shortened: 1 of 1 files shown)


zaquestion:
I'm not sure I understand quite how things work without a config, mostly because I have one. Anyways happy to bring this in, just need to get tests passing. Not sure why its returning with only 1 `/`, but it's something I've seen using `go-gitlab` before.

letientai299:
@zaquestion  please verify again, I've fixed the test.

letientai299:
@zaquestion, so, how do you think about this PR?

PR #313, opened 2019-06-05
Use cobra zsh gen

rsteube:
Sadly the PR concerning zsh completion is still not merged yet, but i managed to extract the changes to a separate package.
This way lab does not need to depend on a cobra fork anymore but instead just has to add an additional dependency (https://github.com/rsteube/cobra-zsh-gen).


rsteube:
The zsh PR is now merged. Opened a new PR for the custom completions.
If that succeeds soon we might as well switch directly to the new version and skip this PR.

zaquestion:
Just seeing this, sounds good, LMK what route you wanna take and we'll do that.

rsteube:
Let's be optimistic for now.

rsteube:
@zaquestion  In the light of https://github.com/spf13/cobra/issues/959 i unarchived https://github.com/rsteube/cobra-zsh-gen and think we should reopen this PR.

Have a look at the `go.mod`, i executed `go mod tidy` and it seems quite some entries were removed by that.

zaquestion:
@rsteube Feel free to reopen (I think you'll need to open new), as long as you merge/rebase in master and do a build with modules enabled it should resolve how the `go.mod` works out.

PR #320, opened 2019-07-25
try XDG_CONFIG_HOME

maxice8:


zaquestion:
I'm open to this and probably merge, but it really makes me think we should just go back to https://github.com/mitchellh/go-homedir -- I glanced over the code and not sure it checks this case, so maybe this change should be proposed over there as well and we could migrate back.

PR #326, opened 2019-08-08
Better ci status

alexkalderimis:
This adds several useful features that I am using on a daily basis:

* allow selecting a trace job by ID
* colours for different statuses
* adding job ID to output so I can easily integrate status and trace commands
* flags for filtering output (only show failures, only show summary)
* nicer summary layout
* display start times and duration
* select the current MR by default in `mr show`

For our workflow, where our pipeline takes over an hour, and has 122 jobs, these features have proved invaluable.

No tests added yet - but since the testdata repo does not have any CI jobs, this is going to be difficult - @zaquestion: could you please add some MRs with CI jobs? Finished ones are fine, ideally with some failed jobs though!

### DIFF
--- a/cmd/ci_status.go
+++ b/cmd/ci_status.go
@@ -7,28 +7,31 @@ import (
 	"text/tabwriter"
 
 	color "github.com/fatih/color"
-        flag "github.com/spf13/pflag"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
+	flag "github.com/spf13/pflag"
+	gitlab "github.com/xanzy/go-gitlab"
 	"github.com/zaquestion/lab/internal/git"
-        gitlab "github.com/xanzy/go-gitlab"
 	lab "github.com/zaquestion/lab/internal/gitlab"
 )
 
 var (
-  onlyFailures bool
-  useColor bool
-  noSkipped bool
-  wait bool
-  noCreated bool
-  summaryOnly bool
-  jobFormat = "%s:\t%s\t-\t%s\tid: %d\n"
-  failed = color.New(color.FgRed)
-  passed = color.New(color.FgGreen)
-  skipped = color.New(color.FgYellow)
-  running = color.New(color.FgBlue)
-  created = color.New(color.FgMagenta)
-  defaultPrinter = color.New(color.FgBlack)
+	onlyFailures   bool
+	useColor       bool
+	noSkipped      bool
+	wait           bool
+	noCreated      bool
+	summaryOnly    bool
+	failed         = color.New(color.FgRed)
+	passed         = color.New(color.FgGreen)
+	skipped        = color.New(color.FgYellow)
+	running        = color.New(color.FgBlue)
+	created        = color.New(color.FgMagenta)
+	defaultPrinter = color.New(color.FgBlack)
+)
+
+const (
+	jobFormat = "%*s: %*s - %10s id: %d\n"
 )
 
 // ciStatusCmd represents the run command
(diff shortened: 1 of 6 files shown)


zaquestion:
@alexkalderimis see the existing CI status tests for how to test with CI jobs, they are in a separate branch in the test data repo, tho I cannot quite remember why I did that anymore ...

https://github.com/zaquestion/lab/blob/master/cmd/ci_status_test.go

internal/gitlab/gitlab.go
```diff
@@ -41,6 +45,10 @@ func User() string {
 	return user
 }
 
+func Client() *gitlab.Client {
```
zaquestion: Let's go ahead and not poke a hole in this package. Any functionality needed from the client should be exposed through this package. Even if said functions, are silly wrappers, still best to stay consistent

internal/gitlab/gitlab.go
```diff
@@ -51,6 +59,72 @@ func Init(_host, _user, _token string) {
 	token = _token
 	lab = gitlab.NewClient(nil, token)
 	lab.SetBaseURL(host + "/api/v4")
+	configDirs := configdir.New("zaquestion", "lab-cli")
```
zaquestion: Is this gonna create a folder on the users filesystem named `zaquestion`?

zaquestion:
It looks like not all of these files have had `gofmt` run, might be wrong since I'm on a phone right now, but please run and update if that's the case.

cmd/ci_status.go
```diff
@@ -21,70 +40,177 @@ var ciStatusCmd = &cobra.Command{
...
+        return fmt.Sprintf("total\tpassed\tfailed\tqueued\n%d\t%d\t%d\t%d",
+                           totalJobs, numPassed, numFailed, numQueued)
+}
+
+func timeMessage(pipeline *gitlab.Pipeline) string {
+        if pipeline.Status == "pending" {
+          return fmt.Sprintf("created at %v", pipeline.CreatedAt)
+        } else if pipeline.Status == "running" {
+          return fmt.Sprintf("started at %v", pipeline.StartedAt)
+        } else {
```
zaquestion: Suggest removed the else's here and above since you are already returning based on the if cases

cmd/ci_status.go
```diff
@@ -21,70 +40,177 @@ var ciStatusCmd = &cobra.Command{
...
+		name = "failures"
+		break
+	}
+	return flag.NormalizedName(name)
+}
+
+func statusColor(status string) *color.Color {
+        switch status {
+        case "failed":
+                return failed
```
zaquestion: Consider moving the var block for these vlaies above this function

cmd/ci_status.go
```diff
@@ -21,70 +40,177 @@ var ciStatusCmd = &cobra.Command{
...
+                return skipped
+        default:
+                return defaultPrinter
+        }
 }
 
 func init() {
+        defaultPrinter.DisableColor()
 	ciStatusCmd.MarkZshCompPositionalArgumentCustom(1, "__lab_completion_remote_branches")
-	ciStatusCmd.Flags().Bool("wait", false, "Continuously print the status and wait to exit until the pipeline finishes. Exit code indicates pipeline status")
```
zaquestion: It's preferred that we either use the above form here or create a config struct specific to this command to house the flags. This is because everything defined at the package level is shared between all of the other commands, and nami g collisions can cause subtle bugs

cmd/ci_status.go
```diff
@@ -21,70 +40,177 @@ var ciStatusCmd = &cobra.Command{
...
+
+        if !summaryOnly {
+          fmt.Fprintln(w, "Stage:\tName\t-\tStatus")
+        }
+        color.NoColor = !useColor
+        var (
+                printer *color.Color
+        )
+        pipelineId := jobs[0].Pipeline.ID
+        pipeline, _, err := lab.Client().Pipelines.GetPipeline(pid, pipelineId)
```
zaquestion: See comment on making these calls functions in the internal GitLab package

internal/gitlab/gitlab.go
```diff
@@ -51,6 +59,72 @@ func Init(_host, _user, _token string) {
...
+	cacheDir = configDirs.QueryCacheFolder()
```
zaquestion: Does this cache stuff actually belong in the GitLab package? Should it be in its own package?

mocha:
@alexkalderimis @leipert do we want to try to finish this up before migrating this over?

bmeneguele:
This PR is stalled for more than a year already. The ideas are valid and can be further enhanced. But since the author didn't replied to any of @zaquestion questions and didn't return with no other PR, I'm going to close it. Feel free to open a new PR in case someone wants to take over the subject.

Many thanks.

PR #327, opened 2019-08-08
Support issue templates

alexkalderimis:
This supports selecting an issue template in the issue create command

There are issues with the git comment character, but that is a git issue.


alexkalderimis:
Now it's my turn to apologise @zaquestion - the formatting has been applied and the branch updated.

Issue #152, opened 2018-05-15
Accept self signed certificates


fauust:
Hi,
is it possible to configure lab to accept self signed certificates?

This is the error:
`mrList.go:38: Get https:/api/v4/projects/<project>: x509: certificate signed by unknown authority`

zaquestion:
@fauust This is something we can probably support. It should be simple enough to allow a configuration option to reference a cert file. At the same time can probably add an option to ignore cert failures.

fauust:
@zaquestion or it is maybe easier to do something link the **wget** `--no-check-certificate` option or
the **curl** `-k, --insecure` option.

tydavis:
I humbly offer my certpool / http client library: https://github.com/tydavis/gobundledhttp

I'm looking into the best way to add to the cert pool on-connect and permanently trust/store it.

zaquestion:
@tydavis looks solid to me, I'm not sure how to use it for self signed certs, but that's largely due to me not being very familiar with actually handling that.

Am I correct in thinking the user would reference a personal cert file?

The insecure client would be immediately handy, we can probably update this function to accept an http client and just pass it through
https://github.com/zaquestion/lab/blob/master/main.go#L65

We'll wanna add a config option, that can just go anywhere for the moment. The config is pretty easy to access with viper. I'll need to think about the overall config structure a bit w.r.t. some other feature requests before any merge. Thoughts appreciated.

darkmattercoder:
What about that issue? Is there a config option for accepting self signed certs? I want to use lab with our corporate instance which uses a self-signed cert.

zaquestion:
@darkmattercoder It should be easy enough to create a config option to allow skipping the certificate check (`--insecure` option mentioned above). Would that work for your purposes?

zaquestion:
@tydavis was working on adding actual support for custom certs, so I had opt'd to see that come in and not add the insecure option, but at this point I don't think that's coming.

darkmattercoder:
Sure, I'm fine with a simple skip.

tydavis:
Sorry -- I never got around to implementing

gaving:
Been burned by this in corporate land, does anybody have a workaround?

zaquestion:
@darkmattercoder @gaving I added config/logic to allow skipping tls verification here: https://github.com/zaquestion/lab/pull/354

Not yet merged, but will bring in and release today after it clears tests.

[...]

PR #359, opened 2019-12-17
Set TLS skip verify when getting user

zaquestion:
Also quick fix to skip user lookup when set to avoid the net call


cwegener:
What is this??? This looks bad!

zaquestion:
@cwegener It's not set by default it's an option you can use in place of loading a self signed cert for gitlab instances that require it. All part of requests in #152

cwegener:
Ah. Yes. I see now. Somehow I got scared off when I first read it. Makes sense now.

Issue #360, opened 2019-12-24
`lab ci status` when local branch name != remote branch name


claytonrcarter:
`lab ci status` has been great (thanks!) but I have recently been working w/ a long running branch that, for a variety of reasons, is names differently locally than on the remote and `lab ci status` doesn't seem to be able to handle this. Does this behavior make sense, or am I misinterpreting something? Does it make sense to try to determine the name of the remote tracking branch and lookup the CI pipelines based on that instead of just the current local branch name?

PR #361, opened 2019-12-24
ci status --wait: actually wait for jobs to finish

claytonrcarter:
This is a quick attempt to address the issues in #240, namely that `lab ci status --wait` is a *wee* bit too generous w/ it's interpretation of the word "wait". Basically, this moves the fetching of the CI jobs into the `for` loop so that it's fetching fresh status for every iteration. (I added a 1s delay between loop runs in case API rate limits are an issue. Not really sure, but though it was worth it to be careful and kind.)

I've marked this as WIP because I don't know if/how this affects complex, multi-pipeline jobs and such. All of my jobs are very simple, with just a single job and I don't have anything more complex to test against.

Also, I did some misc cleanup of the code while I was in there, and I added a more informative error message to as an interim step to properly addressing #360


claytonrcarter:
Hmm, the build is failing w/
```
GO111MODULE=on go test -coverprofile=coverage-main.out -covermode=count -coverpkg ./... -run= github.com/zaquestion/lab/cmd github.com/zaquestion/lab/internal/...
--- FAIL: Test_fork (2.91s)
    fork_test.go:64: failed to delete project during cleanup: DELETE https://gitlab.com/api/v4/projects/15958339: 500 {message: 500 Internal Server Error}

FAIL
```

Not sure what I did (or if I did anything) to trigger a 500 error from GL.

claytonrcarter:
Hmm, something else I'm seeing is that the formatting is slightly off between the first iteration of the loop and the rest:

```
Stage: Name - Status
test:  test - running

test: test - running

test: test - running

...
```

Note the extra space in `test:  test` on the first line. I think it has something to do w/ the `fmt.Fprintln(w)` affecting the tab width or something, but now we're getting out of my depth. If I move that call to `fmt.Fprintln(w)` around to different locations, it definitely affects how much space there is between `test:` and `test`. 🤷‍♂

zaquestion:
hmm I'm surprised to see this change, I'm not sure why I ever thought it would have worked.....

Tests: Could be a gitlab issue, they be testing on prod all the time

Regarding the spacing, that just has to do with the tab writer. 
These are lined up because they aren't separated by a newline, which is what
```
Stage: Name - Status
test:  test - running
```
`fmt.Fprintln(w)` is just trying to add a blank newline, from the tab writers perspective subsequent `test:  test - running` don't need to align. This is fine, when the pipelines have more than 1 job they all get printed in a block and align with cycle.

claytonrcarter:
OK, 👍 for the tab issue. That makes sense. I've kicked off another build to see if it's better. I've been using this more or less daily for a few weeks and it has been solid. Happy to make any changes or updates you'd like, as usual.

claytonrcarter:
😕 still getting the 500 from `fork_test.go:64`

zaquestion:
@claytonrcarter Another open PR is seeing the same failure, must be a change on GL side, I'm hoping to investigate a bit further this weekend. Defs not related to the changes in here.

claytonrcarter:
I removed the "WIP" from the title since I guess this more or less good.

zaquestion:
I guess the tests are fine now :shrug:

Issue #318, opened 2019-07-11
Markdown preview of the merge requests text and comments


XVilka:
Since lab allows to view the list of MRs, I think it will be next step level experience to support markdown preview of the PR text (most of them are text anyway) and attached comments.
Same applies to the issues.

And there is a markdown preview console application in Go already: https://github.com/samfoo/mdcat

It can even be made scrollable, like in [termimad](https://github.com/Canop/termimad):
![image](https://user-images.githubusercontent.com/203261/61017333-659a3300-a3c5-11e9-994f-468930a8a916.png)

zaquestion:
@XVilka whoa nice, I've thought about this before, but never considered there would be a solid solution. Never even tried to google it... Would love to get this implemented where applicable

PR #362, opened 2020-01-13
Adding glamour support for issue descriptions

craftamap:
Greetings!

[charmbracelet/glamour](https://github.com/charmbracelet/glamour) is a go commandline markdown renderer.

This pull request shows a simple implementation of glamour into the issue_show function. If you pass the flag ´-g´ onto ´lab issue show 18´, glamour will be used to render the markdown.


zaquestion:
This looks promising! I believe there's at least 1 issue we can link that desires something like this. If we can get the tests and ideally a screen shot up I see no reason not to bring this in as the default 💪

craftamap:
sure, I will take a look into this the next days!

muesli:
Linking related issue: WIP: #318

craftamap:
although the last test failed, if you take a look at the log, all test concerning this pr don't throw errors anymore.
I have a few questions about glamour-rendering:
 
Right now, only the dark theme of glamour is used in the pr. This however could be bad for persons using a light theme as their setup. While glamour supports an auto-light/dark-mode, this only allows switching between the default dark and light themes. 

The default themes have some specific characteristics to them: 
 - an empty line before and after the actual content is rendered
 - two spaces are rendered before every line
If we do not want to use those settings, we have to define our own styles. While this is not much of an issue, we would have to implement the auto-mode for dark/light on our side.

What do you think about this? @zaquestion 

I will also attach screenshots on how it looks right now.

craftamap:
image

muesli:
Author of glamour here, just a few remarks from my side:

>       If we do not want to use those settings, we have to define our own styles. While this is not much of an issue, we would have to implement the auto-mode for dark/light on our side.

You could also retrieve the auto-style (`AutoStyleConfig`) from glamour and adjust only some values of it (like the indentation, suffixes & prefixes you want it to use). That means you don't have to reimplement any of the terminal background detection logic.

zaquestion:
@craftamap I'm comfortable with the leading spacing, I actually think the way it looks in your snapshot is pretty clean with contents indented beneath the title. I imagine it'll make reading issues a little easier when scrolling back within the terminal.

That said, it looks like @muesli has suggested a suitable way to address this. I'm comfortable with either option

zaquestion:
I'll try and figure out what's up with the tests this weekend, excited to bring this in.

PR #368, opened 2020-03-02
updated shell completion

### LINKED ISSUE #333: Zsh completion doesn't show "repo" alias for "project"
It looks like it was solved in the upstream (https://github.com/spf13/cobra/pull/669) but your version is pinned at `v0.0.0-20180412120829-615425954c3b` (I'm not a gopher, so i'm not 100% sure that that's the reason)

### LINKED ISSUE #367: Add fish completions
Would be nice to provide `fish` shell completions in addition to `bash` and `zsh`

rsteube:
Had another go at what i originally extracted from the cobra zsh PR: https://github.com/rsteube/cobra-zsh-gen
This brings lab back to cobra (no fork) and makes the completion less cryptic as it's all go code.

- adds dependency to https://github.com/rsteube/cobra-zsh-gen
- adds dependency to https://github.com/src-d/go-git (seemed resonable to me as this makes git handling a bit easier than invoking the shell, but can change that if it's a bad idea)

Needs some good testing (looks ok so far).

fixes #333 
fixes #367


zaquestion:
`go-get` dependency is fine, in the long term I'd like to move everything over, I got many of the calls working a while back even https://github.com/zaquestion/lab/tree/go-git -- it's honestly considerably faster than exec'ing the binary.

zaquestion:
Tests are kinda fucky right now due to some changes on gitlab.com, I'm going to go ahead and merge this on faith since you've consistently delivered in the past. Also if you're interested in gaining contributor access so you can bring these through a little easier I'm super open to that.

cmd/action/action.go
```diff
@@ -0,0 +1,78 @@
+package action
```
zaquestion: This package seems like it would be more appropriately place under `internal` -- `cmd` is really for the commands themselves.

zaquestion:
Actually, ideally lets move the action package into `internal` and then we can bring this in.

cmd/action/action.go
rsteube: yeah, wasn't too sure where to place it - moved it to `internal/action`

rsteube:
What does contributor access do? I am rather worried if means i can accidentally push to the wrong remote XD.

rsteube:
I managed to adapt this for `fish` and `bash` as well which i merged into a new project: https://github.com/rsteube/carapace
API is the same (just different package name) and what i tested looks pretty good so far so i'm updating this MR with it.

zaquestion:
@rsteube Looks like the PR I just merged caused a conflict in this. Can you update and we'll get it brought in?

zaquestion:
Contributor access would let you merge, I don't think anyone can push to master, but I would need to double check the controls

rsteube:
Ok, thats fine then - i probably have to to fix some small quirks anyway when people start using it.

zaquestion:
@rsteube I've added you as a collaborator. Not sure the state of the tests, so I (or someone) may need to fix those, ideally you'll be able to merge this on your own and bring in updates at your discretion

zaquestion:
I can see the ci stuff is pretty borked at this point, and I introduced some conflicts bringing in a few other changes. Let's get the conflicts resolved and I'll bring this in and work to fix the pipelines so we can release

rsteube:
Nice - yeah at least at one point the url simply changed. I think it was sth like an added slash another "breadcrumb" in the repository url.

rsteube:
@zaquestion seems the conflict was just an additonal import.
"Fixed" two tests, but seems the go-gitlab dependency needs an update: https://github.com/xanzy/go-gitlab/blob/master/project_snippets.go#L94
`Code` is renamed to `Content` (https://docs.gitlab.com/ee/api/project_snippets.html#create-new-snippet)

Apart from that the build itself is ok - shall i do a rebase first?

rsteube:
Ok, seems updating that wasn't too hard. But now https://gitlab.com/lab-testing/test/-/merge_requests/1385 already exists, so the test fails.

rsteube:
moved ci fixes to separate PR: https://github.com/zaquestion/lab/pull/386

zaquestion:
@rsteube I think I got the travis stuff running again, I've gotta wait for a the checks to run on a PR before I can make them required, go ahead and try updating this branch to the latest they should run

zaquestion:
Verfied the tests are passing with your fixes as well https://travis-ci.org/github/zaquestion/lab/builds/719466634

PR #375, opened 2020-03-30
readme: warn before aliasing

glensc:
the aliasing is very dangerous, it can kill your terminal:

image

the problem is that my shell prompt executes git commands to draw prompt. yet lab is pretty aggressive about it's configuration. maybe it's problem on it's own that it should ask for gitlab credentials ONLY when these are needed, proxying things if they are missing.

I've debugged and found the command that my prompt used to be:

```
[~/scm/tmp/castnow (master)★] ➔ alias git='_git() { echo >&2 "git $*"; git "$@"; }; _git'
[~/scm/tmp/castnow (master)★] ➔ cd .
git rev-parse @
[~/scm/tmp/castnow (master)★] ➔
```

```
[~/scm/tmp/castnow (master)★] ➔ lab rev-parse @
Enter GitLab host (default: https://gitlab.com): ^C
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,9 +122,17 @@ fpath=(/path/to/script_directory $fpath)
 
 # Aliasing
 
-Like hub, lab feels best when aliased as `git`. In your `.bashrc` or `.bash_profile`
+Like hub, lab feels best when aliased as `git`. In your `.bashrc` or `.bash_profile`:
+
 ```
 alias git=lab
+```
+
+NOTE: before aliasing, if you use git in your shell prompt command, be sure lab works by it's own first:
+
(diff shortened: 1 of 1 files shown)


zaquestion:
I'll bring this in since it's generally good advice. Really it's best when defining prompts that use git to use `_git` for all since it won't be aliased and you can avoid the problem altogether. Of course in the long wrong if we ever make it to a 1.0 I'd like to remove the recommendation to alias altogether and instead use git aliases (https://git-scm.com/book/en/v2/Git-Basics-Git-Aliases) to surface the "git extension" feel

Issue #338, opened 2019-10-03
lab ci view does not render if "only: merge_requests"  a.k.a merge_request_pipelines is used 


lukmdo:
`lab ci view` does not render merge_request_pipelines (ones using `only: merge_requests`) 
https://docs.gitlab.com/ee/ci/merge_request_pipelines/

```
$ lab ci view
2019/10/03 19:10:54 ci_view.go:484: <nil>

$ lab version 
git version 2.22.0
hub version 2.12.1
lab version 0.16.0
```

I created two branches 
1. https://gitlab.com/lukmdo/test/tree/test_merge_request_pipelines corresponding to MR 
https://gitlab.com/lukmdo/test/merge_requests/1
which generates two pipelines 
- branch pipeline https://gitlab.com/lukmdo/test/pipelines/86375805
- merge requests pipeline labeled as `detached` https://gitlab.com/lukmdo/test/pipelines/86375806
when run `lab ci view` this only renders `branch pipeline`  
<img width="199" alt="Screen Shot 2019-10-03 at 7 22 29 PM" src="https://user-images.githubusercontent.com/87109/66153599-bcee1f00-e613-11e9-95cd-97b80d8c0e14.png">

that is ~OK however **expected** latest (merge requests pipeline in this case) would be rendered

2. https://gitlab.com/lukmdo/test/tree/check_only_merge_request_pipelines corresponding to MR 
https://gitlab.com/lukmdo/test/merge_requests/2 
which generates only 
- merge requests pipeline labeled as `detached` https://gitlab.com/lukmdo/test/pipelines/86378820
when run `lab ci view` this fails 
```
$ lab ci view
2019/10/03 19:10:54 ci_view.go:484: <nil>
```
**expected** rendered merge requests pipeline with two jobs 

See all pipelines https://gitlab.com/lukmdo/test/pipelines

zaquestion:
I suspect this has to do with how lab selects the pipeline to render, before `merge_request` pipelines you could get away with only rendering the branch pipeline because there was no difference (and the same pipeline was displayed on the MR anyhow). 

Here's what the pipelines look like from your test repo (thanks for putting that together)
```
[
  {
    "id": 86378820,
    "sha": "2710a09d5ed85b1414c5cee5af50f3bd587bd9f5",
    "ref": "refs/merge-requests/2/head",
    "status": "success",
    "created_at": "2019-10-03T18:10:22.873Z",
    "updated_at": "2019-10-03T18:12:31.483Z",
[...]

PR #380, opened 2020-06-26
base changes for issue #338

lukmdo:
For issue https://github.com/zaquestion/lab/issues/338

## main changes 
- use of `sha` instead of `ref/branchName` https://docs.gitlab.com/ee/api/pipelines.html#list-project-pipelines is more correct with detached pipelines ( via `only: merge_requests` )
- use `sha` of upstream (not local)

## additionally  
- add `GetBranch` and remove/replace `BranchPushed`  
- `cmd` parse args unification
  - use `parseArgsRemoteString`
  - add `parseArgsRemoteRef` (helper)

### DIFF
--- a/cmd/ci_view.go
+++ b/cmd/ci_view.go
@@ -24,8 +24,9 @@ import (
 )
 
 var (
-	projectID int
-	branch    string
+	projectID  int
+	branchName string
+	commitSHA  string
 )
 
 // ciViewCmd represents the ci command
@@ -49,15 +50,15 @@ Feedback Encouraged!: https://github.com/zaquestion/lab/issues`,
 			remote string
 			err    error
 		)
-		branch, err = git.CurrentBranch()
+		branchName, err = git.CurrentBranch()
 		if err != nil {
 			log.Fatal(err)
 		}
 
 		if len(args) > 1 {
-			branch = args[1]
+			branchName = args[1]
 		}
-		remote = determineSourceRemote(branch)
+		remote = determineSourceRemote(branchName)
 		if len(args) > 0 {
 			ok, err := git.IsRemote(args[0])
 			if err != nil || !ok {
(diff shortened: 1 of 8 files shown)


cmd/ci_view.go
```diff
@@ -77,6 +78,11 @@ Feedback Encouraged!: https://github.com/zaquestion/lab/issues`,
 			log.Fatal(err)
 		}
 		projectID = project.ID
+		branch, err := lab.GetBranch(projectID, branchName)
```
lukmdo, line 81: this needs redo as before `view [remote [branch/tag]]` was accepting both as ref not only  branchName

I am thinking of sth like 
```golang
func FindCommitByRef(pid interface{}, ref string) (*gitlab.Commit, error) {}
```
via `lab.Commits.ListCommits` https://docs.gitlab.com/ce/api/commits.html#list-commits

lukmdo: done on https://github.com/zaquestion/lab/pull/380/commits/468bb6916394a58f32ee962b98cdbdbbefac920f

lukmdo:
@zaquestion help with review - hope you will like this

cmd/root_test.go
```diff
@@ -404,6 +404,73 @@ func Test_parseArgsRemoteString(t *testing.T) {
 	}
 }
 
+func Test_parseArgsRemoteRef(t *testing.T) {
```
lukmdo, line 407: not too DRY (based on Test_parseArgsRemoteString) - but should be still OK ?

zaquestion:
@lukmdo this looks good, I originally implemented the ci jobs stuff as sha and switched away since I ran into issues retrieving the latest pipelines on reruns or something like that. I think the issues I hit were things like generally not seeing the jobs/pipelines you'd expect if you were viewing the CI on the MR or in the UI.

That said, your implementation looks stronger than what I originally crapped out, so I'm inclined to bring this in and see how it goes.

zaquestion:
Oh now I see, and remember that we already discussed a lot of this in https://github.com/zaquestion/lab/issues/338

PR #391, opened 2020-08-15
mr create: Set default target branch

prarit:
The default branch on gitlab projects can be changed from master to other
values.  The 'mr create' command defaults to 'master' and should use the
project's default branch instead.

Use the project's default branch instead of 'master'.

### DIFF
--- a/cmd/mr_create.go
+++ b/cmd/mr_create.go
@@ -26,7 +26,7 @@ var mrCreateCmd = &cobra.Command{
 	Use:     "create [remote [branch]]",
 	Aliases: []string{"new"},
 	Short:   "Open a merge request on GitLab",
-	Long:    `Creates a merge request (MR created on origin master by default)`,
+	Long:    `Creates a merge request (default: MR created on default branch of origin)`,
 	Args:    cobra.MaximumNArgs(2),
 	Run:     runMRCreate,
 }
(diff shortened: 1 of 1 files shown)


prarit:
@zaquestion , not sure what to do to pass these.  I'm fixing a bug.  One thought is that perhaps a new test repo could be added that has a "main" branch instead of "master"?  FWIW I think that's overkill ;)

P.

cmd/mr_create.go
```diff
@@ -124,14 +124,15 @@ func runMRCreate(cmd *cobra.Command, args []string) {
 	if err != nil {
 		log.Fatal(err)
 	}
-	targetBranch := "master"
+
+	targetBranch := lab.GetDefaultBranch(targetProjectName)
```
zaquestion, line 128: We probably don't need to add a function for this, we already have the target projects so:
```suggestion
	targetBranch := targetProject.DefaultBranch
```

zaquestion:
@prarit The tests passed, the failures you're seeing are due to lack of coverage. They aren't strictly required to merge, but this does callout that this codepath isn't properly exercised which is concerning, because I would have expected https://github.com/zaquestion/lab/blob/master/cmd/mr_test.go#L27 to hit it....

zaquestion:
Also thanks for this change, in addition to making lab more flexible, it'll general enable more inclusive default branch names 👍

cmd/mr_create.go
prarit: I'll make this change now and retest.

PR #393, opened 2020-08-18
note: Add mr note --file option

prarit:
Add the ability to update a merge request with the contents of a file.
This is especially useful when copying in testing contents and/or
boot logs.

Add mr note --file option which takes the location of the file containing
the text as an argument.

### DIFF
--- a/cmd/mr_note.go
+++ b/cmd/mr_note.go
@@ -8,6 +8,7 @@ import (
 	"strconv"
 	"strings"
 	"text/template"
+	"io/ioutil"
 
 	"github.com/rsteube/carapace"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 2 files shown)


cmd/mr_note.go
```diff
@@ -97,6 +112,7 @@ func mrNoteText() (string, error) {
 
 func init() {
 	mrCreateNoteCmd.Flags().StringSliceP("message", "m", []string{}, "Use the given <msg>; multiple -m are concatenated as separate paragraphs")
+	mrCreateNoteCmd.Flags().StringP("file", "f", "", "Use the given file as the message")
```
zaquestion, line 115:
```suggestion
	mrCreateNoteCmd.Flags().StringP("file", "F", "", "Use the given file as the message")
```
I'm leaning towards the short flag being a `-F` since that's how the comparable flag works on `git commit`, which is really the syntax being followed for a lot of these commands that.

zaquestion: What are your thoughts? honestly I could go either way since `-f` is pretty common for this usecase too.

prarit: That's a good point about git -F.  I'll change this to -F and retest.

PR #401, opened 2020-08-23
.gitignore: Add internal go files

prarit:
Ignore go.sum and go.mod.  These are metadata files for go and can
be safely ignored.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,10 @@ _test
 coverage*
 vendor
 
+# go metadata
+go.sum
+go.mod
+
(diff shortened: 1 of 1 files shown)


zaquestion:
We should be committing these https://github.com/golang/go/wiki/Modules#should-i-commit-my-gosum-file-as-well-as-my-gomod-file

zaquestion:
That said, using `go mod tidy` can be key to avoid committing in things brought in during testing or by mistake.

prarit:
Oh interesting.  I haven't been committing my go.sum and go.mod files.  I will definitely add that to my list.

Issue #402, opened 2020-08-23
config file usage?


prarit:
Hey @zaquestion,

There are a few features for which I will need to use the configfile.  I'm not sure if there any restrictions on using it ATM and want to get clarification for it's use.

Here are some of the things I'm thinking of using the config file for.  Please note my examples are in a gitconfig format and I do understand that the lab configfile is in hcl.  The gitconfig format just seemed easier to type :)

a) Storing gitlab personal access tokens

Instead of storing them in text I think the way to really do this is to implement something like

[token]
gitlab.com = pass show gitlab/token

so that lab could have multiple gitlab instance's tokens stored.

b) Alias'ing commands

Similar to git-alias, the config file would be a place to store lab alias commands. For example, instead of a 'git mr show --patch' command (which would create the patches for a merge request), an alias of something like

[alias]
mr-show-patch = !sh -c 'mr checkout;  git-format-patch'

keeps the amount of code we have to worry about to a minimum.

c) Highlighting new comments

One of the features that is lost with an email based git workflow (ie the kernel) is the ability to keep track of new comments.  In a mailer this is easy because the new comments remain unread but in a text-based console it's not so easy to do.

I've been toying around with a 'mr show --since' option that highlights "new" merge request comments [1] .  For example, 'lab mr show 9 --comments --since "2020-08-21 00:08:15.318 +0000 UTC"' would only show comments that were added since Aug 21 @ 8:15 AM.  It's actually easy to code using github.com/fatih/color.

After implementing --since I realized that I had created a bad user experience as users have to remember the date  they last viewed an MR or Issue.  After thinking about it some more a better approach is to keep some metadata (that can be flushed every so often so it doesn't get out-of-hand) that simply stores the last time an MR or Issue was displayed on the console.  The lab command would read the stored value and highlight new comments since that date.

This would have to be opt-in as I can definitely see not every user caring about highlighting of new comments.

So something like

[show-metadata]

gitlab.com/project-name mr 8 = 2020-08-21 14:57:46.808 +0000 UTC
gitlab.com/project-name mr 200 = 2020-08-20 11:12:12.511 +0000 UTC
gitlab.com/project-name issue 12 = 2020-08-16 09:45:11.612 +0000 UTC

The data could be cleared for CLOSED/MERGED issues (on linux via systemctl timers, cron, or atd) nightly or through a command line option, 'lab clean-metadata'.

[1] https://github.com/prarit/lab/tree/mr_show_since

[...]

PR #405, opened 2020-08-26
show mr and show issue automatic highlight logging

prarit:
Implement on-demand and automatic new comment detection on MRs and Issues.

P.

### DIFF
--- a/cmd/issue_show.go
+++ b/cmd/issue_show.go
@@ -3,6 +3,7 @@ package cmd
 import (
 	"fmt"
 	"log"
+	"os"
 	"strings"
 	"time"
 
@@ -11,6 +12,7 @@ import (
 	"github.com/fatih/color"
 	"github.com/rsteube/carapace"
 	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
 	gitlab "github.com/xanzy/go-gitlab"
 	"github.com/zaquestion/lab/internal/action"
 	lab "github.com/zaquestion/lab/internal/gitlab"
@@ -53,7 +55,7 @@ var issueShowCmd = &cobra.Command{
 				log.Fatal(err)
 			}
 
-			printDiscussions(discussions, since)
+			printDiscussions(discussions, since, int(issueNum))
 		}
 	},
 }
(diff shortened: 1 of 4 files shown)


prarit:
@zaquestion , given that this is a visible user change I thought it is best to get your opinion on it before merging.

To test pick an MR (I'm assuming #9 below) and o

lab mr show 9 --comments
lab mr note 9 
# add a new note
lab mr show 9 --comments

which should highlight the new comment.

P.

cmd/mr_show.go
```diff
@@ -98,7 +106,43 @@ WebURL: %s
...
+	sinceIsSet := false
+	CompareTime, err = dateparse.ParseLocal(since)
+	if err != nil || CompareTime.IsZero() {
+		CompareTime = viper.GetTime(mrEntry)
+		if CompareTime.IsZero() {
+			CompareTime = time.Now().UTC()
+		}
+	} else {
+		sinceIsSet = true
+	}
```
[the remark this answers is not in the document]
prarit: Ha :)  I rewrote that section of code *twice*, and yours still looks better than mine :)

As for the .git directory: 

> @prarit Sorry for the delay I responded to the larger questions over on #402
> 
> For this change, I'm a tad hesitant to write this data into `.git` directories, but I'm more than open to storing the metadata somewhere. Usually I just aim to come up with some kind of adequate for what we do in user space and "git" space, so if you have a good case for putting `labmetadata` in `.git/` I could be swayed.
> 
> In #402 I suggest leveraging `~/.cache` or `~/.local`, but I realize that's a bit more machinery, and it's a lot easier to look for a file relative to the repo root. That said, things written into `.git/` are ignored by default, so it's not the worst move to write there...

I will also respond there.

P.

cmd/mr_show.go
```diff
@@ -98,7 +106,43 @@ WebURL: %s
 		mr.Author.Username, milestone, labels, mr.WebURL)
 }
 
-func printMRDiscussions(discussions []*gitlab.Discussion) {
+func printMRDiscussions(discussions []*gitlab.Discussion, since string, mrNum int) {
+	NewAccessTime := time.Now().UTC()
+	viper.Reset()
+	mrEntry := fmt.Sprintf("mr%d", mrNum)
+	viper.SetConfigName("labmetadata")
+	viper.SetConfigType("hcl")
```
zaquestion, line 114: As alluded to in #402 this doesn't need to be hcl, long term we want to move away from that for the other config as well. Go ahead and use whatever format you feel makes sense here.

cmd/mr_show.go
```diff
@@ -98,7 +106,43 @@ WebURL: %s
...
+	viper.AddConfigPath(".git")
+	// write data
+	if _, ok := viper.ReadInConfig().(viper.ConfigFileNotFoundError); ok {
+		if err := viper.WriteConfigAs(".git/labmetadata.hcl"); err != nil {
```
zaquestion, line 118: As mentioned in other comment, doesn't need to be hcl, but specifically here, I'm on board to put files in `.git` (particularly since these are ignored by git by default and being local to the repo solves a lot of problems), but I'd like to talk a bit more up front on some naming conventions.

First I'd like to go with either a `lab_` preview to file names or place files in a `.git/lab/` directory (leaning towards the later). Second, does `metadata` make sense to describe whats being stored in this file? It's certainly metadata, but it's also specific to `lab issue/mr show` operations, so something describing that seems preferred. Perhaps `.git/lab/show_op_log` or `.git/lab/cmd_show_metadata` or `.git/lab/show_metadata` (with w/e extension)

cmd/mr_show.go
prarit: I'm going to stick with hcl for now just for the sake of consistency.   I'll invest some time in looking at other alternatives.

cmd/mr_show.go
prarit: I like the last option.  .git/lab/show_metadata sounds good to me.

cmd/issue_show.go
```diff
@@ -108,7 +116,43 @@ WebURL: %s
...
+	sinceIsSet := false
+	CompareTime, err = dateparse.ParseLocal(since)
+	if err != nil || CompareTime.IsZero() {
+		CompareTime = viper.GetTime(issueEntry)
+		if CompareTime.IsZero() {
+			CompareTime = time.Now().UTC()
+		}
+	} else {
+		sinceIsSet = true
+	}
```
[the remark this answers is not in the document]
prarit: Will be fixed in new version.

prarit:
> Eeeep, didn't mean to approve, please see outstanding comments

np.  I've just pushed a version that addresses the code change you've requested as well as moving the config file to .git/lab/show_metadata.hcl

I'll wait for a final sign off from you before merging.

PR #409, opened 2020-08-31
(ci) Initial implementation of GitLab CI

leipert:


.gitlab-ci.yml
```diff
@@ -0,0 +1,60 @@
...
+    # Temporarly create Ci for every branch:
+    - if: '$CI_COMMIT_BRANCH'
+    # For tags, create a pipeline.
+    - if: '$CI_COMMIT_TAG'
+    # For stable branches, create a pipeline.
+    - if: '$CI_COMMIT_BRANCH =~ /^[\d-]+-stable$/'
+    #
+
+default:
+  image: golang:1.13
```
zaquestion, line 16: We can update the version we test to `1.15` and `1.14`, generally I just test the last 2 versions, but have been slacking on updating the existing CI
```suggestion
  image: golang:1.15
```

leipert: @zaquestion I will add 1.15 😄 The README still describes 1.13 as the minimal requirement, and I thought it might be best to not change it all in one go (pun intended) 😆

zaquestion: Eh, in thinking about this more, I'd only test 1.15 (the version we'll use to build the release), but we can also verify that `1.14` is able to build the project. We can update the readme to set `1.14` as the minimum.

zaquestion:
The tests largely use lab with this configuration https://github.com/zaquestion/lab/blob/master/testdata/lab.hcl

zaquestion:
The tests failing on gitlab may have to do with these steps missing from CI https://github.com/zaquestion/lab/blob/master/.travis.yml#L23-L29

I should have some time to look closer later this week. Unfortunately until Thurs, I'm "enjoying" my first week oncall at the day job.

zaquestion:
I take it back `401 {message: 401 Unauthorized}` isn't related to the ssh config, it's the test binary picking up the CI credentials with it's "automatic CI config" mode. Hmmm, need to figure out a way to tell lab to use the correct creds when running in it's own pipeline. 😆

Issue #411, opened 2020-09-01
Config file overwritten if user is empty


bmeneguele:
Hi @zaquestion and @prarit,

In commit 1285411b8584bac827bf876e6c1d44180e871d5b @prarit forced the `core.user` config option to be written to the config file in case it's empty, however Viper doesn't update/merge the configuration file, but overwrites it: if you have a config file like the following one:
```
"core" = {
  "host" = "https://gitlab.com"
  "user" = ""
  "token" = "XXXXXXXXXXXXXX"
}
```
And you run `lab`, the code messes up with the config file:
```
"core" = {
  "user" = "bmeneg"
}
```
`viper.WriteConfigAs()` only writes those options that got `viper.Set()`'ed, all the others are not considered.

With that,
- should we really touch the config file from within the app without user's knowledge? Wouldn't it be better to warn the user to set the `core.user` option instead?

I know you guys are planning to overhaul the configuration stuff soon, but I think we should fix it regardless of the rework - unless you think it'll be released pretty soon :-D.

Another thing, this same behavior would be interesting to be seen in all other options: what happens if `core.user` is there but `core.host` and `core.token` aren't?
```
2020/09/01 18:47:07 main.go:92: <nil>
2020/09/01 18:47:07 main.go:93: missing config value core.host in /home/bmeneg/.config/lab.hcl
exit status 1
```
A suggestion I think would be nice is: 
- separate between which options are optional and which are required
  - if the config file exists, but some required options are missing: ask the user to enter it from stdin (like it's done when the config file doesn't exist.
  - if the config file exists, but some optional options are missing: do nothing.

What do you guys think?
I can help with getting it fixed while the you guys can focus on the overhaul.

zaquestion:
In the short term, it might be easiest to just revert that commit. Originally user was in the config, but I removed it to make lab "smarter" and remove 1 less step for the user. That decision has caused more heartache than I would have hoped for...

In any case, ideally we'd lookup the user around https://github.com/zaquestion/lab/blob/master/main.go#L69 so it's written in when the config is created. I'm not sure why I didn't implement it that way in the first place, so it may be some doing to make that happen.

prarit:
Ouch.  @bmeneguele, can you revert 1285411b8584 ("config: Fix missing core.user entry") locally and test?  If that works I'll revert ASAP.

I can reproduce this locally and reverting that commit seems to resolve the problem.  I will revert after you confirm.

bmeneguele:
[...]

PR #412, opened 2020-09-02
Revert commit 1285411b8584 ("config: Fix missing core.user entry")

prarit:
lab commit 1285411b8584 attempted to fix the missing core.user entry in
~/.config/lab.hcl, however, it seems to have caused more problems than
it fixed [1].

I'm reworking the config code and have noticed several other odd issues so
reverting this isn't a major step backward.

Reported-by: Bruno Meneguele <bmeneg@redhat.com>

[1] https://github.com/zaquestion/lab/issues/411

### DIFF
--- a/main.go
+++ b/main.go
@@ -132,9 +132,8 @@ func loadConfig() (string, string, string, bool) {
 	}
 	if user == "" {
 		user = getUser(host, token, tlsSkipVerify)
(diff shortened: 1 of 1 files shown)


prarit:
@zaquestion, it just occurred to me that the config tests in main_test.go should have failed after I applied this patch.

PR #414, opened 2020-09-02
New config layout

prarit:
WIP: New config layout

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -2,8 +2,11 @@ package config
 
 import (
 	"bufio"
+	"bytes"
 	"fmt"
 	"io"
+	"io/ioutil"
+	"log"
 	"net/url"
 	"os"
 	"strings"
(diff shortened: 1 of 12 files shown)


main.go
```diff
@@ -56,92 +65,35 @@ func loadConfig() (string, string, string, bool) {
...
-	// is tough to simplify since cfg["host"] is accessing the array "core"
-	// and viper.GetString("core.host") is expecting a non-array so it
-	// doens't match
-	if v := viper.GetString("core.host"); v != "" {
-		cfg["host"] = v
-	}
-	if v := viper.GetString("core.token"); v != "" {
-		cfg["token"] = v
+	user = getUser(host, token, tlsSkipVerify)
+	if os.Getenv("LAB_CORE_TOKEN") != "" || os.Getenv("LAB_CORE_HOST") != "" {
```
bmeneguele: I'm a bit confused by this part: why core.user is being set only when LAB_CORE_TOKEN or LAB_CORE_HOST are set?

main.go
```diff
@@ -56,92 +65,35 @@ func loadConfig() (string, string, string, bool) {
...
+		viper.Set("core.user", user)
+		viper.WriteConfig()
```
bmeneguele: I see that using viper.WriteConfig() here don't replace all options not viper.Set()'ed like it did when using HCL format. Was it also an issue with the format?

prarit:
> I see that using `viper.WriteConfig()` here don't replace all options not `viper.Set()`'ed like it did when using HCL format. Was it also an issue with the format?

Oh, that's a bug then.  If viper.WriteConfig() doesn't write out all the config then yes, I have to use viper.WriteConfigAs().  I'll double check the behaviour.

main.go
prarit: I'm just writing the core.user value because it's new.  I'll double check that this works as expected and writes the whole config.

main.go
prarit: Yup, I see what you're getting at and I fixed this.  FWIW, I hadn't yet tested setting LAB_CORE_TOKEN or LAB_CORE_HOST to see if that chunk of code worked.  I've tested it now ;) and the result is a small code change.

main.go
prarit: Same thing as the previous comment.  I've fixed this so that it works when LAB_CORE_TOKEN or LAB_CORE_HOST is set.  I've verified that the entire config is rewritten.

zaquestion:
I don't know why codecov is commenting every 5 lines on the stuff that isn't covered by tests ... Super noisy, sorry folks I try to find out how to disable that (unless it's just me?)

internal/config/config.go
```diff
@@ -86,3 +94,151 @@ func CI() (string, string, string) {
...
+	viper.SetConfigType("toml")
+	viper.WriteConfigAs(newconfig)
+
+	// delete the old config HCL file
+	err = os.Remove(oldconfig)
+	if err != nil {
+		fmt.Println("Warning: Could not delete old config file", oldconfig)
+	}
+
+	// HACK
```
zaquestion, line 128: fwiw, it doesn't surprise me at all that we had to do this. Should be fine since lab configs have remained simple, essentially for this reason too.

zaquestion (commented):
This is looking great already, I should have time tomorrow to weigh in more and catch up on the other stuff. Thanks for taking this on, huge improvement, and I never expected we'd migrate the only configs so that's a nice bonus

internal/config/config.go
prarit: > I don't know why codecov is commenting every 5 lines on the stuff that isn't covered by tests ... Super noisy, sorry folks I try to find out how to disable that (unless it's just me?)

Hmm ... I'm by no means an expert in codecov.  We do have a few engineers who are familiar with it.  Let me see if I can ping someone to take a look.

bmeneguele:
Nicely done @prarit !
I've tested all cases I could think of and things seem to be working pretty well.

prarit:
@zaquestion, I've dropped WIP from the description and the patches.  This code is ready for review IMO.

zaquestion:
I'll be doing a deep dive over the weekend, thanks @prarit !

bmeneguele (approved):
AFAICT the changes are working as expected and the code is pretty sane and clean.
Thanks @prarit .

zaquestion:
Looking solid, I took it for a spin locally and saw my configs convert seamlessly. I added a simple test for the convert func, after the tests pass lets move forward and bring this in.

zaquestion (approved):

zaquestion:
@prarit I'll let you do the honors 👍

PR #416, opened 2020-09-04
Load token

prarit:


prarit:
@zaquestion , there's some subtlety of the config testing that I'm not quite understanding.  Could you review my changes here and maybe make a suggestion as to why you think they're failing?

prarit:
> @zaquestion , there's some subtlety of the config testing that I'm not quite understanding. Could you review my changes here and maybe make a suggestion as to why you think they're failing?

Actually, never mind.  I figured it out.

internal/config/config.go
```diff
@@ -161,6 +180,26 @@ func getUser(host, token string, skipVerify bool) string {
 	return u.Username
 }
 
+// GetToken returns a token string from the config file.
+// The token string can be cleartext or returned from a password manager or
+// encryption utility.
+func GetToken() string {
+	token := viper.GetString("core.token")
+	if token == "" && viper.GetString("core.load_token") != "" {
+		_token, err := exec.Command(viper.GetString("core.load_token")).Output()
```
bmeneguele, line 189: I tried to run it with GPG (decrypting a file containing the token, somewhat how I do with my email clients) and it failed because `exec.Command()` requires the actual command args to be passed as a separate argument: `func Command(name string, arg ...string) *Cmd` from os/exec godoc.

Otherwise the error I get is this one:
'```2020/09/08 18:22:14 config.go:191: fork/exec gpg2 --no-tty -q -d /home/bmeneg/gitlab-token.gpg: no such file or directory```

It seems to consider the whole command name + its args as a single `/usr/bin/'gpg2 --no-tty..'` file.

bmeneguele (changes requested):

internal/config/config.go
prarit: Oof :)  @bmeneguele always breaking my code :)  Let me hack something together for you to test.

prarit: @bmeneguele, I did something dumb and didn't comment on the code before forcing pushing. Sorry 'bout that.  Here's the code in question:


                   `_token, err := exec.Command(viper.GetString("core.load_token")).Output()`

and I changed it to

```
                // args[0] isn't really an arg ;)
                args := strings.Split(viper.GetString("core.load_token"), " ")
                _token, err := exec.Command(args[0], args[1:]...).Output()
```

ie) I split the string into an executable part and a args part.  This has a minor drawback in that if someone did "SET_RANDOM_ENV_VAR=foo /tmp/load_token.sh -f -g -r -123" then it wouldn't work and would error out.  I'm hoping no one goes that far ;)

prarit: @bmeneguele in any case, let me know if it works for you and I can merge.  Thanks :)

internal/config/config_test.go
```diff
@@ -176,3 +232,47 @@ func TestConvertHCLtoTOML(t *testing.T) {
   user = "lab-testing"
 `, string(cfgData))
 }
+
+func TokenTest(t *testing.T) {
```
zaquestion, line 236: Tests need to start with be in the form `TestXxx` to run properly. 

Ref: https://golang.org/pkg/testing/

zaquestion (changes requested):
Need to fix the test names so they actually run, should help with the coverage warning too.

internal/config/config_test.go
zaquestion: Here and for `TokenLoadTest`

prarit: Ah thanks -- testing and fixing.

internal/config/config.go
bmeneguele: Yea, I don't think it's worth checking user input for env vars.. we can set it's required to be exported from bashrc or something like that if the user wants it.

For the change, it's working like a charm now! :+1:
Many thanks @prarit .

bmeneguele (approved):
The MR is fine including the changes on `load_token` handling both app name and args separated, plus the test naming fix.

internal/config/config_test.go
prarit: Okay fixes applied.  Thanks for catching that and the pointer to testing.

zaquestion (approved):

PR #417, opened 2020-09-06
Mr show config comment

prarit:
Add a local "comments" config option that always sets "--comments" for the mr and issue show commands.

### DIFF
--- a/cmd/mr_show.go
+++ b/cmd/mr_show.go
@@ -24,6 +24,9 @@ var (
 	mrShowPatchReverse bool
 )
 
+// default path for metadata config file
+var metadatafile = ".git/lab/show_metadata.toml"
+
 var mrShowCmd = &cobra.Command{
 	Use:        "show [remote] <id>",
 	Aliases:    []string{"get"},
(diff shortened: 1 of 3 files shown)


cmd/mr_show.go
```diff
@@ -41,6 +44,27 @@ var mrShowCmd = &cobra.Command{
 			log.Fatal(err)
 		}
 
+		viper.Reset()
+		viper.AddConfigPath(".git/lab")
+		viper.SetConfigName("show_metadata")
+		viper.SetConfigType("toml")
+		// write data
+		if _, ok := viper.ReadInConfig().(viper.ConfigFileNotFoundError); ok {
```
zaquestion, line 52: Might be good to abstract some of this config handling into the config package, perhaps some general purpose functionality for command specific config. It also seems a tad strange that the config overrides (like `comments = true`) can only be set in the repo specific config?

prarit: It's funny that you bring this up.  When I was writing the code I was thinking the exact same thing: this code could all be centralized into the config code but wasn't 100% sure if you wanted it there.  I'll definitely do that in another merge.

prarit: > Might be good to abstract some of this config handling into the config package, perhaps some general purpose functionality for command specific config. It also seems a tad strange that the config overrides (like `comments = true`) can only be set in the repo specific config?

I moved this code over into the config in https://github.com/zaquestion/lab/pull/430.  As for the config overrides :) I'm working on getting the overrides working.  I'm pretty close and hopefully should have something in the next few days for your review.

Issue #374, opened 2020-03-28
lab --version requires api authentication


nijel:
Invoking with simple args such as ``--version`` or ``--help`` should not require authentication to GitLab and should not query the remote server.

Without configuration:

```
lab --version
Enter GitLab host (default: https://gitlab.com): 
```

When hitting rate limiting on GitLab:

```
lab --version
2020/03/28 19:36:45 main.go:178: GET https://gitlab.com/api/v4/user: 429 failed to parse unknown error format
```

prarit:
Looks like this was resolved in #421

PR #421, opened 2020-09-08
main: skip config initialization for --version and --help

bmeneguele:
Resolves issue #374 by skipping config initialization for `--version` and `--help` options passed directly to `lab` command.

### DIFF
--- a/main.go
+++ b/main.go
@@ -33,6 +33,10 @@ func skipInit() bool {
 		return false
 	}
 	switch os.Args[1] {
+	case "--version", "version":
+		return true
+	case "--help", "help":
+		return true
 	case "completion":
 		return true
 	case "_carapace":


main.go
```diff
@@ -33,6 +33,10 @@ func skipInit() bool {
 		return false
 	}
 	switch os.Args[1] {
+	case "--version":
+		return true
+	case "--help":
```
zaquestion, line 38:
```suggestion
	case "--help", "help":
```
Probably worth adding the non-flag variant as well.

zaquestion (approved):

main.go
bmeneguele: Good point.
It seems that "version" also has a 'non-flag' variant as well. I just pushed a new version of the patch with both added.

prarit: Changes LGTM.

Issue #422, opened 2020-09-08
--no-markdown|-M flags with no action on parent command


bmeneguele:
--no-markdow,-M are described in the parent `lab mr` and `lab issue` commands, however it's only used in the `show` subcommand of these, which have their own implementation and description in place. With that, I think this flag should be removed from these parent commands to avoid confusion, like I had while trying to use it.

PR #423, opened 2020-09-08
issue,mr: remove markdown flag from parent command

bmeneguele:
Resolves issue #422 that reports confusion in the -M|--no-markdown flag usage in parent commands.


zaquestion (approved):
nice catch

Issue #425, opened 2020-09-09
MR Description doesn't display correctly


prarit:
I've noticed this a few times when using lab   If I execute 'lab mr create' and enter text like

```
Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1876934
Upstream Status: merged into the linux.git
Brew: https://brewweb.engineering.redhat.com/brew/taskinfo?taskID=31161820
Tested: manually, with reproducers, see bz1876934 #c5, #c7

commit 7cf64b18b0b96e751178b8d0505d8466ff5a448f
Author: Eric Biggers <ebiggers@google.com>
Date:   Sat Mar 21 20:43:05 2020 -0700

    vt: vt_ioctl: fix use-after-free in vt_in_use()
```
Then gitlab will mangle the first 4 lines into a single paragraph and either mark the "^commit" line as the start of code, or the "vt: vt_ioctl" line as the start of code.

This then translates poorly into 'lab mr show' which displays

```
Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1876934 Upstream Status: merged into the linux.git
Brew: https://brewweb.engineering.redhat.com/brew/taskinfo?taskID=31161820 Tested: manually, with reproducers, see bz1876934 #c5, #c7

commit 7cf64b18b0b96e751178b8d0505d8466ff5a448f
Author: Eric Biggers <ebiggers@google.com>
Date:   Sat Mar 21 20:43:05 2020 -0700

    vt: vt_ioctl: fix use-after-free in vt_in_use()
```

instead of the expected first section above.

Normally I would say that we should tell contributors to go fix up their MR Description after the submission, however, since most (all?) of us are using a text editor on the command line perhaps we should default to wrapping the MR Description with ``` ?

@zaquestion, any thoughts here?  I would completely understand if you said "Leave it as is" FWIW.  I'm definitely focused on making this easy for contributors using text editors ;)

prarit:
```
diff --git a/cmd/mr_create.go b/cmd/mr_create.go
index 4000411f760e..e976e10b0d31 100644
--- a/cmd/mr_create.go
+++ b/cmd/mr_create.go
@@ -179,6 +179,10 @@ func runMRCreate(cmd *cobra.Command, args []string) {
                }
        }
 
+       // This is text output so wrap the body in 'code' tags to avoid any
+       // issues with formatting.
+       body = "```"+body+"```"
+
[...]

PR #429, opened 2020-09-10
lab: Add --force-linebreak option

prarit:
FIXME: needs testing fixes
FIXME: needs better option and text

The default gitlab formatting will mangle some types of text by default.

For example, if the text entered as the Description contains

This is sentence one and should not be
On the same line as sentence two

then gitlab would merge these lines together as

This is sentence one and should not be On the same line as sentence two.

To fix this add an option to replace line breaks, "\n", with "  \n" [1]

[1] https://gist.github.com/shaunlebron/746476e6e7a4d698b373.

### DIFF
--- a/cmd/issue_create.go
+++ b/cmd/issue_create.go
@@ -65,6 +65,11 @@ var issueCreateCmd = &cobra.Command{
 			log.Fatal("aborting issue due to empty issue msg")
 		}
 
+		linebreak, _ := cmd.Flags().GetBool("force-linebreak")
+		if linebreak {
+			body = textToMarkdown(body)
+		}
+
 		assigneeIDs := make([]int, len(assignees))
 		for i, a := range assignees {
 			assigneeIDs[i] = *getAssigneeID(a)
@@ -138,6 +143,7 @@ func init() {
 	issueCreateCmd.Flags().StringSliceP("label", "l", []string{}, "Set the given label(s) on the created issue")
 	issueCreateCmd.Flags().StringSliceP("assignees", "a", []string{}, "Set assignees by username")
 	issueCreateCmd.Flags().StringP("template", "t", "default", "use the given issue template")
+	issueCreateCmd.Flags().Bool("force-linebreak", false, "append 2 spaces to the end of each line to force markdown linebreaks")
 
 	issueCmd.AddCommand(issueCreateCmd)
 	carapace.Gen(issueCreateCmd).PositionalCompletion(
--- a/cmd/util.go
+++ b/cmd/util.go
@@ -1,0 +1,13 @@
+// This file contains common functions that are shared in the lab package
+package cmd
+
+import (
+	"strings"
+)
+
+// textToMarkdown converts text with markdown friendly line breaks
+// See https://gist.github.com/shaunlebron/746476e6e7a4d698b373 for more info.
+func textToMarkdown(text string) string {
+	text = strings.Replace(text, "\n", "  \n", -1)
+	return text
+}
--- a/cmd/util_test.go
+++ b/cmd/util_test.go
@@ -1,0 +1,14 @@
+package cmd
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func Test_textToMarkdown(t *testing.T) {
+	basestring := "This string should have two spaces at the end."
+	teststring := basestring + "\n"
+	newteststring := textToMarkdown(teststring)
+	assert.Equal(t, basestring+"  \n", newteststring)
+}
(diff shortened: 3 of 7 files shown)


prarit:
@zaquestion, second attempt at resolving https://github.com/zaquestion/lab/issues/425.  Please note that while this doesn't completely capture my brain ATM it does accurately describe my difficulty in selecting an option name and help text:

https://i.redd.it/2gw6jt7ju8m51.jpg

cmd/issue_create.go
```diff
@@ -138,6 +143,7 @@ func init() {
 	issueCreateCmd.Flags().StringSliceP("label", "l", []string{}, "Set the given label(s) on the created issue")
 	issueCreateCmd.Flags().StringSliceP("assignees", "a", []string{}, "Set assignees by username")
 	issueCreateCmd.Flags().StringP("template", "t", "default", "use the given issue template")
+	issueCreateCmd.Flags().Bool("rawtext", false, "Do not use markdown renderer for text")
```
zaquestion, line 146: `force-linebreak` comes to mind

zaquestion: or even `--br` named after the html for it.

zaquestion: The help text here should capture what this feature is doing, less what it's trying to undo, which hopefully will make naming the feature and writing a description easier.

`"append 2 spaces to the end of each line to force markdown linebreaks"` <- even an off the cuff description makes it pretty clear what using the flag is going to do. Just my 0.002 cents

internal/gitlab/gitlab.go
```diff
@@ -888,3 +888,10 @@ func Labels(labels []string) *gitlab.Labels {
 	l := gitlab.Labels(labels)
 	return &l
 }
+
+// TextToGitlabText converts text with Gitlab friendly line breaks.
+// See https://gist.github.com/shaunlebron/746476e6e7a4d698b373 for more info.
+func TextToGitlabText(text string) string {
```
zaquestion, line 894: This isn't really gitlab specific behavior, it's just markdown, so on the topic of naming I don't think this is named correctly, nor in the right package. I don't really have a better suggestion tho, should of just being an unexported function in the `cmd` package...

https://markdown-guide.readthedocs.io/en/latest/basics.html#line-return

prarit: Y'know when I add the --comment flag to 'mr show' I noticed that the 'issue show' and 'mr show' commands had VERY similar structure and was wondering if there was a common place for lab code.  I would have unified some of the 'mr show' and 'issue show' code.

I haven't seen one?  If there isn't one perhaps creating "cmd/common.go" is a good idea.  I'm sure there are other functions that could be made common.

What do you think?  The code base is eventually going to need a common area.

prarit: I just pushed a new version with a cmd/common.go file (and the other fixes discussed above) to see what you think.

zaquestion: To date, I've put common stuff in `cmd/root.go`, but we probably should have made a `cmd/common.go` or `cmd/util.go` a while ago.

prarit: Let me look at root.go and think about util.go.  I do think a util.go is the way to go.

prarit: I've pushed a util.go and made the other changes.  Let me know what you think.  I'll put in a test for util.go as well.

cmd/util_test.go
```diff
@@ -0,0 +1,13 @@
+package cmd
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TesttextToGitlabText(t *testing.T) {
```
zaquestion, line 9:
```suggestion
func Test_textToGitlabText(t *testing.T) {
```
This won't run without an uppercase following `Test`, however `_` counts for this.

prarit: Fixed.

cmd/util.go
```diff
@@ -0,0 +1,13 @@
+// This file contains common functions that are shared in the lab package
+package cmd
+
+import (
+	"strings"
+)
+
+// textToGitlabText converts text with markdown friendly line breaks
+// See https://gist.github.com/shaunlebron/746476e6e7a4d698b373 for more info.
+func textToGitlabText(text string) string {
```
bmeneguele, line 10: As @zaquestion mentioned before, that's not a specific GitLab behavior, that's basically Markdown's. Wouldn't it be better to name this function something like "textToMdText" or call it LF (line feed as in ASCII table) "toMdLF()" to make it clear that the function only handles \n? I don't know, just thinking out loud here :-D

prarit: Yup, moved to cmd/util.go and added a test for it too.

prarit: Pushed new version with changes.

bmeneguele (approved):
Unfortunately, Markdown still messes with some portions of the plain text formatting, but the patch solves the issue for sure :) Thanks @prarit !

Issue #376, opened 2020-04-06
StringSliceP vs. StringArrayP usage for flags


odise:
Hi,

I just came around an issue while adding notes to project issues like the one below:

```
lab issue note 1 -m "{\"key\": \"value\"}"
Error: invalid argument "{\"key\": \"value\"}" for "-m, --message" flag: parse error on line 1, column 3: bare " in non-quoted-field
```

After a bit of research I found that this is related to the usage of `pflags` StringSliceP parameter type used on various places in the code. Specifically `GetStringSlice` is raising the error above. 

I just would like to put https://github.com/spf13/pflag/issues/157 as a reference here and suggest using `StringArrayP` flag type at least for the issue command part. I'm unsure about the others which is the reason for raising the issue without a PR ready to solve it. 

Any thoughts from your end?

zaquestion:
Interesting, at the moment I don't recall making a conscious choice to choose one over the other, and I'm thinking I probably thought they were alaises to the same functionality... 

Thanks for the reference issue, seeing the difference clearly, I think `StringArrayP` seems more appropriate for a lot of our stuff I imagine. Particularly the message flags. Things like assignees and labels SliceP seems more fitting. Just a gut take, PRs welcome, if we change the message flag would just like to see that be consistent across the various objects.

prarit:
Since #431 was merged I think this can be closed.

PR #431, opened 2020-09-15
flags: fix "message" string handling

bmeneguele:
Solves issue #376 wrt string handling in `--message | -m` flags.

### DIFF
--- a/cmd/mr_note_test.go
+++ b/cmd/mr_note_test.go
@@ -11,18 +11,42 @@ import (
 )
 
 func Test_mrCreateNote(t *testing.T) {
-	repo := copyTestRepo(t)
-	cmd := exec.Command(labBinaryPath, "mr", "note", "lab-testing", "1",
-		"-m", "note text")
-	cmd.Dir = repo
-
-	b, err := cmd.CombinedOutput()
-	if err != nil {
-		t.Log(string(b))
-		t.Fatal(err)
+	tests := []struct {
+		Name         string
+		Args         []string
+		ExpectedBody string
+	}{
+		{
+			Name:         "Normal text",
+			Args:         []string{"lab-testing", "1", "-m", "note text"},
+			ExpectedBody: "https://gitlab.com/lab-testing/test/merge_requests/1#note_",
+		},
+		{
+			// Escaped sequence text direct in the argument list as the
+			// following one was already a problem:
+			// https://github.com/zaquestion/lab/issues/376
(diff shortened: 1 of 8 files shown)


prarit:
@bmeneguele 

This looks good, however, I think adding 'message' tests for the commands that explicitly test message input like "{\"alpha\": \"beta\"}" is a good idea.  After all, that is the bug you're fixing.  Take a look at cmd/mr_note_test.go:Test_mrNoteMsg() for example.  You could easily add a new test to that file that tests the example input.

bmeneguele:
Yes, of course. It makes huge sense in adding a test for it.
Thanks for the heads up @prarit  :P

zaquestion (approved):
Looks great, not that adding a test is a bad move either ;) 

@bmeneguele I've sent an invite to add you as a collaborator as well :D Thanks for all the great work to improve lab.

bmeneguele:
> @bmeneguele I've sent an invite to add you as a collaborator as well :D Thanks for all the great work to improve lab.

I'm glad I can help! 
Tyvm @zaquestion

bmeneguele:
I added the test for the string causing issues, but it had to follow a different path from what you suggested @prarit. After checking the test source I noticed that it was testing the resulting note msg being created, giving no room for checking the msg coming from the argument list (during app exec).

The issue happens right in the argument list parsing, so I added the test by calling the 'lab' with the problematic text and checked its output. What do you think?

cmd/mr_note_test.go
```diff
@@ -23,6 +23,19 @@ func Test_mrCreateNote(t *testing.T) {
...
+	cmd = exec.Command(labBinaryPath, "mr", "note", "lab-testing", "1",
+		"-m", "{\"key\": \"value\"}")
+
+	b, err = cmd.CombinedOutput()
+	if err != nil {
+		t.Log(string(b))
+		t.Fatal(err)
+	}
+
+	require.Contains(t, string(b), "https://gitlab.com/lab-testing/test/merge_requests/1#note_")
```
zaquestion, line 38: Suggest wrapping this in a subtest, and linking to the issue ( https://github.com/zaquestion/lab/issues/376 )

bmeneguele: Sure thing. Will do.
Thanks!

bmeneguele:
@zaquestion I finally added the subtest for the problematic string and also linked it to the issue number.
Sorry for vanishing for some days :)

PR #432, opened 2020-09-15
Adds creation of Merge Request discussions

marwatk:
This adds the ability to create threads on a Merge Request. It's basically a clone of the note command but talks to the Discussions API instead.

It's very simple but can be expanded later. We use automated discussions on all MRs for things like "Did you update the README?".

I couldn't make the tests pass on my system since we're on go 1.14 and `testing.T.TempDir` isn't available, but it looks like CI is present so I'm hoping to watch it pass there.

### DIFF
--- a/cmd/mr_discussion.go
+++ b/cmd/mr_discussion.go
@@ -78,7 +78,7 @@ func mrDiscussionMsg(msgs []string) (string, error) {
 	if err != nil {
 		return "", err
 	}
-	return git.EditFile("MR_NOTE", text)
+	return git.EditFile("MR_DISCUSSION", text)
 }
 
 func mrDiscussionText() (string, error) {
--- a/cmd/mr_discussion.go
+++ b/cmd/mr_discussion.go
@@ -1,0 +1,122 @@
+package cmd
+
+import (
+	"bytes"
+	"fmt"
+	"io/ioutil"
+	"log"
+	"runtime"
+	"strconv"
+	"strings"
+	"text/template"
+
+	"github.com/rsteube/carapace"
+	"github.com/spf13/cobra"
+	gitlab "github.com/xanzy/go-gitlab"
+	"github.com/zaquestion/lab/internal/action"
+	"github.com/zaquestion/lab/internal/git"
+	lab "github.com/zaquestion/lab/internal/gitlab"
+)
+
+var mrCreateDiscussionCmd = &cobra.Command{
+	Use:     "discussion [remote] <id>",
+	Aliases: []string{"comment"},
+	Short:   "Start a discussion on an MR on GitLab",
+	Long:    ``,
+	Args:    cobra.MinimumNArgs(1),
+	Run: func(cmd *cobra.Command, args []string) {
+		rn, mrNum, err := parseArgs(args)
+		if err != nil {
+			log.Fatal(err)
+		}
+
+		msgs, err := cmd.Flags().GetStringSlice("message")
+		if err != nil {
+			log.Fatal(err)
+		}
+
+		filename, err := cmd.Flags().GetString("file")
+		if err != nil {
+			log.Fatal(err)
+		}
+
+		body := ""
+		if filename != "" {
+			content, err := ioutil.ReadFile(filename)
+			if err != nil {
+				log.Fatal(err)
+			}
+			body = string(content)
+		} else {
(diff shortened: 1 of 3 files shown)


cmd/mr_discussion.go
```diff
@@ -0,0 +1,122 @@
...
+func mrDiscussionMsg(msgs []string) (string, error) {
+	if len(msgs) > 0 {
+		return strings.Join(msgs[0:], "\n\n"), nil
+	}
+
+	text, err := mrDiscussionText()
+	if err != nil {
+		return "", err
+	}
+	return git.EditFile("MR_NOTE", text)
```
zaquestion, line 81:
```suggestion
	return git.EditFile("MR_DISCUSSION", text)
```

marwatk: Thanks, thought I caught all those.

marwatk: [Fixed](https://github.com/zaquestion/lab/pull/432/commits/e8b79e8dc84714ffbd7ed0d8ad3b91536ede9f86)

marwatk:
If you'd like me to refactor note and discussion to share more code let me know.

zaquestion:
I partly wonder if discussions should just be a flag atop the existing note functionality. More of what's on my mind tho is how does one reply to a discussion? Is it possible to do that we the current changes?

bmeneguele:
> I couldn't make the tests pass on my system since we're on go 1.14 and testing.T.TempDir isn't available, but it looks like CI is present so I'm hoping to watch it pass there.

`t.TempDir()` is used only in `internal/config/config_test.go` while your changes is in `cmd`, so you skip the internal-related tests by running:

```
$ GO111MODULE=on go test ./cmd
```

marwatk:
> I partly wonder if discussions should just be a flag atop the existing note functionality.

Initially I was going to do just that (because I'm lazy and it would have been easier) but the [discussion API](https://docs.gitlab.com/ce/api/discussions.html#create-new-merge-request-thread) has _way_ more features than the [note API](https://docs.gitlab.com/ee/api/notes.html#create-new-merge-request-note) that could be expanded later and wouldn't apply to notes at all, so I figured it should be its own thing.

> More of what's on my mind tho is how does one reply to a discussion? Is it possible to do that with the current changes?

Not yet, but it would easy to expand it later to add something like a `--reply-to 1234` or something that calls `AddMergeRequestDiscussionNote` in `go-gitlab`.  If you think there's value in it now I can definitely flesh it out a bit more.

marwatk:
I spent the morning trying to implement reply, but to me the best way would be something like:
```
lab mr reply <mr id> <discussion id>
```

But the only parsing functions allowing for the optional remote only accommodate up to two arguments total (including the optional remote). I noticed the comment that you'd like to clean that up, and I started (and had what I thought was good progress) when I hit this case:
```
Use:     "list [remote] [search]",
```

My refactor falls apart there because there are two optional positional arguments and I'm not sure that can be properly handled. What if, for example, someone wanted to search for `origin`? It would get picked up as a remote instead of a search. I spent some time trying to accommodate two optional positional arguments but everything ended up just as messy as the current stuff.

I suppose I could just implement my own command line parsing in the `reply` command, but that seems silly. It didn't feel right to add more mess to the parsing to accommodate multiple positional arguments. And finally it didn't feel right to make one of the requisite arguments a flag instead.

So I punted and I'm going to leave this MR as-is unless you're ok with me converting the `search` param to a flag on `list`, which would be a breaking change.

zaquestion:
Yeah all the different `parseArgs` functions got out of hand real fast. That said theres no requirement that every command uses them, they are only broken out like that, since it was such a common pattern for the commands. These changes are good as is, anything else we can build on in the next PR, thanks for all of your thoughtful efforts here ❤️

prarit:
> I spent the morning trying to implement reply, but to me the best way would be something like:
> 
> ```
> lab mr reply [remote] <mr id> <discussion id>
> ```
> 

lab mr show <mr_id> --patch functionality uses findLocalRemote() which IMO might be useful here?  That way you can drop the [remote] arg, and just use 'lab mr reply <mr_id> <discussion_id>

I have an idea on how to show the discussion IDs in the the output.  See https://github.com/zaquestion/lab/commit/8b0cc20ffbcb342d6a1fda9075c9ea09c3839cd5 which might be useful for you.

Issue #364, opened 2020-02-10
lab mr checkout 39: mr_checkout.go:33: the key `remote.origin.url` is not found


moltar:
I'm trying to checkout a MR:

```sh
lab mr checkout 39
```

But my checkout doesn't have an `origin` remote:

```sh
❯ git remote -v
github	git@www-github.foo.com:org/repo.git (fetch)
github	git@www-github.foo.com:org/repo.git (push)
gitlab	git@192.16.50.1:org/repo.git (fetch)
gitlab	git@192.16.50.1:org/repo.git (push)
```

How to deal with cases like this?

Thanks!

zaquestion:
You should be able to pass the remote as an argument

`lab mr checkout gitlab 39`

zaquestion:
If that doesn't work probably an easy fix

moltar:
That doesn't work, unfortunately :)

```
❯ lab mr checkout gitlab 47
Error: accepts 1 arg(s), received 2
Usage:
  lab mr checkout [flags]

Flags:
  -b, --branch string   checkout merge request with <branch> name
  -h, --help            help for checkout
  -t, --track           set checked out branch to track mr author remote branch, adds remote if needed
```

PR #433, opened 2020-09-16
mr_checkout: fix usage message

bmeneguele:
This merge request is pretty simple, only adds the '<id>' text to the `lab mr checkout -h` command.
No issue opened for that.

### DIFF
--- a/cmd/mr_checkout.go
+++ b/cmd/mr_checkout.go
@@ -25,7 +25,7 @@ var (
 
 // listCmd represents the list command
 var checkoutCmd = &cobra.Command{
-	Use:   "checkout",
+	Use:   "checkout <id>",
 	Short: "Checkout an open merge request",
 	Long:  ``,
 	Args:  cobra.ExactArgs(1),
--- a/cmd/mr_checkout.go
+++ b/cmd/mr_checkout.go
@@ -28,7 +28,7 @@ var checkoutCmd = &cobra.Command{
 	Use:   "checkout <id>",
 	Short: "Checkout an open merge request",
 	Long:  ``,
-	Args:  cobra.ExactArgs(1),
+	Args:  cobra.MinimumNArgs(1),
 	Run: func(cmd *cobra.Command, args []string) {
 		rn, mrID, err := parseArgs(args)
 		if err != nil {


bmeneguele:
Wait a sec, just noticed issue #364, let me work on that as well.

cmd/mr_checkout.go
```diff
@@ -25,7 +25,7 @@ var (
 
 // listCmd represents the list command
 var checkoutCmd = &cobra.Command{
-	Use:   "checkout",
+	Use:   "checkout <id>",
 	Short: "Checkout an open merge request",
 	Long:  ``,
 	Args:  cobra.ExactArgs(1),
```
zaquestion, line 31: Now that I can see it infront of me, fixing #364 should just be a matter of:
```suggestion
	Args:  cobra.MinimumNArgs(1),
```

bmeneguele: I was thinking in also add the option the handle the "remote" positional argument.
Not sure how desirable it is.. what do you think?

zaquestion: Yeah, that's the kinda sad part with this command, it seems to already have support for it, it's just the argument constraint that messes it up.

```
	rn, mrID, err := parseArgs(args)
		rn, mrID, err := parseArgs(args)
		if err != nil {
		if err != nil {
			log.Fatal(err)
			log.Fatal(err)
		}
		}
		mrs, err := lab.MRList(rn, gitlab.ListProjectMergeRequestsOptions{
		mrs, err := lab.MRList(rn, gitlab.ListProjectMergeRequestsOptions{
			IIDs: []int{int(mrID)},
			IIDs: []int{int(mrID)},
		}, 1)
```

bmeneguele:
One thing that I noticed is the checkout misbehaving with the [remote] option when set.
I'm trying to figure out what is actually going on or what I've done wrong. Tomorrow I may have something on that (even though just the reproducer).

bmeneguele:
> One thing that I noticed is the checkout misbehaving with the [remote] option when set.
> I'm trying to figure out what is actually going on or what I've done wrong. Tomorrow I may have something on that (even though just the reproducer).

Will open an issue specific for that..

PR #434, opened 2020-09-17
Config rework

prarit:
The lab config heirarchy is:
    1. ENV variables (LAB_CORE_TOKEN, LAB_CORE_HOST)
            - if specified, core.token and core.host values in
              config files are not updated.
    2. "dot" . user specified config
            - if specified, lower order config files will not override
              the user specified config
    3.  .config/lab/lab.toml (global config)
    4.  .git/lab/lab/toml (worktree config)

Values from the worktree config will override any global config settings.

PR #436, opened 2020-09-20
Merge Flags and Configs

prarit:
Merge the flags and configs so that a user can override any config setting on the command line.  For example if the lab config file contained

[mr_foo]
  clioption = false

then by default, 'lab mr fool' would have clioption 'false' unless '--clioption true' was specified.

### DIFF
--- a/cmd/mr_checkout.go
+++ b/cmd/mr_checkout.go
@@ -25,10 +25,11 @@ var (
 
 // listCmd represents the list command
 var checkoutCmd = &cobra.Command{
-	Use:   "checkout <id>",
-	Short: "Checkout an open merge request",
-	Long:  ``,
-	Args:  cobra.MinimumNArgs(1),
+	Use:              "checkout <id>",
+	Short:            "Checkout an open merge request",
+	Long:             ``,
+	Args:             cobra.MinimumNArgs(1),
+	PersistentPreRun: LabPersistentPreRun,
 	Run: func(cmd *cobra.Command, args []string) {
 		rn, mrID, err := parseArgs(args)
 		if err != nil {
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -228,6 +228,29 @@ func parseArgsRemoteRef(args []string) (string, string, error) {
 	return rn, name, nil
 }
 
+// setCommandPrefix returns a concatenated value of some of the commandline.
+// For example, 'lab mr show' would return 'mr_show.', and 'lab issue list'
+// would return 'issue_list.'
+func setCommandPrefix(scmd *cobra.Command) {
+	for _, command := range RootCmd.Commands() {
+		if CommandPrefix != "" {
+			break
+		}
+		commandName := strings.Split(command.Use, " ")[0]
+		if scmd == command {
+			CommandPrefix = commandName + "."
+			break
+		}
+		for _, subcommand := range command.Commands() {
+			subCommandName := commandName + "_" + strings.Split(subcommand.Use, " ")[0]
+			if scmd == subcommand {
+				CommandPrefix = subCommandName + "."
+				break
+			}
+		}
+	}
+}
+
 var (
 	// Will be updated to upstream in Execute() if "upstream" remote exists
 	forkedFromRemote = "origin"
@@ -314,6 +337,10 @@ func Execute() {
 			return
 		}
 	}
+
+	// Set CommandPrefix
+	scmd, _, _ := cmd.Find(os.Args)
+	setCommandPrefix(scmd)
 
 	if err := RootCmd.Execute(); err != nil {
 		// Execute has already logged the error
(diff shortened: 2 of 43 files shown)


prarit:
This explicitly depends on the changes in https://github.com/zaquestion/lab/pull/434

prarit:
@zaquestion the codecov failures aren't as bad as they seem.  They occur because I'm adding a line of code to every command, and technically I'm only testing one command in my tests (mr show --comments).  I think that's enough FWIW to get adequate testing of the flags and configs merge code.

prarit:
rebased for #434 and to separate out testing changes from last patch.

cmd/root.go
```diff
@@ -315,6 +315,26 @@ func Execute() {
...
+			break
+		}
+		for _, subcommand := range command.Commands() {
+			subCommandName := commandName + "_" + strings.Split(subcommand.Use, " ")[0]
+			if scmd == subcommand {
+				CommandPrefix = subCommandName + "."
+				break
+			}
+		}
+	}
```
zaquestion, line 336: This looks involved enough to merit breaking into a function and unittesting specifically.

zaquestion (commented):
Overall this is surprisingly straightforward, would like to let this sit open for at least a few more days, so that there's ample opportunity for interested parties to look at it or test drive before it's merged.

zaquestion:
> @zaquestion the codecov failures aren't as bad as they seem. They occur because I'm adding a line of code to every command, and technically I'm only testing one command in my tests (mr show --comments). I think that's enough FWIW to get adequate testing of the flags and configs merge code.

You must have resolved this, codecov is showing 94% diff coverage, and the tests you've added mostly seem to cover the new functionality

cmd/util.go
```diff
@@ -14,26 +16,58 @@ var (
...
+		)
+
+		switch f.Value.Type() {
+		case "bool":
+			configValue = getMainConfig().GetBool(CommandPrefix + f.Name)
+			configString = strconv.FormatBool(configValue.(bool))
+		case "string":
+			configValue = getMainConfig().GetString(CommandPrefix + f.Name)
+			configString = configValue.(string)
+		case "stringSlice":
```
zaquestion, line 36: may need to include support for `stringArray` (not sure if it's a unique type under the hood...), per https://github.com/zaquestion/lab/pull/431

prarit: I think I'm okay because I have a log.Fatal() for the default case.  IOW code goes BOOM if there's an unidentified type.

cmd/root.go
prarit: I've been thinking about this comment for a bit.  The problem with unit testing is figuring out a way to fake a Command and subCommand structure and confirming the values that come out of it are correct.  I wonder if a better way is to add something like

if CommandPrefix != "mr_show." {
     log.Fatal("CommandPrefix", CommandPrefix", does not match "mr_show.")
}

throughout the test files?  That's a pain to maintain though.

prarit: I'll definitely move this into a function.

cmd/mr_checkout.go
```diff
@@ -25,10 +25,11 @@ var (
...
 // listCmd represents the list command
 var checkoutCmd = &cobra.Command{
-	Use:   "checkout <id>",
-	Short: "Checkout an open merge request",
-	Long:  ``,
-	Args:  cobra.MinimumNArgs(1),
+	Use:              "checkout <id>",
+	Short:            "Checkout an open merge request",
+	Long:             ``,
+	Args:             cobra.ExactArgs(1),
```
zaquestion, line 31: This should go back to 
```
cobra.MinimumNArgs(1)
```

cmd/root.go
zaquestion: You don't wanna do a `log.Fatal` in the tests, to say the least, as it'll kill coverage output. There's `t.FailNow()` IIRC which if it doesn't accept a log statement can be preceded by `t.Log()`. Yeah, agreed tho, recreating the command structure is a bit annoying and a unittest won't catch if things move around inside cobra. 

Looks like you've already broken it out into a function now, so I'll call this thread resolved. I'm not concerned so much with unittesting as you've covered the functionality with the integration tests.

cmd/mr_checkout.go
prarit: Oops.  Good catch.  Will fix.

cmd/util.go
prarit: So LOL after a rebase the code did go boom.  The problem is that viper doesn't have support for stringArray.  For now I've put a comment in the code that stringArray is unsupported in viper.  Given how stringArray is being used in the code with -m on note commands, I doubt anyone will want to set it in the config files.  It's something we can fix later.

cmd/mr_checkout.go
prarit: Pushed with new version.  Sorry 'bout that.

Issue #438, opened 2020-09-25
hardcoded remote discovery for working tree commands


bmeneguele:
There are situations where the clone or fork process isn't done via the lab tool, but the user wants to use it regardless, thus the remote being used by the user may not follow the pattern that the lab tool uses: `upstream`, `origin` or `lab.User()`.
```
$ git remote -v
origin-2        git@gitlab.com:bmeneg/Test-Lab.git (fetch)
origin-2        git@gitlab.com:bmeneg/Test-Lab.git (push)
$ lab mr list
2020/09/25 17:45:11 mr_list.go:36: the key `remote.origin.url` is not found
```
However, some commands accept the `[remote]` positional arg:
```
$ lab mr list origin-2
#2 fix test file
```
Those commands that perform remote operations (based on the Gitlab library) they work based on the project name, so they are fine most of the cases (didn't test all of them).
However, those commands that interact with the local/working tree get lost:
```
$ lab mr checkout 2
2020/09/25 17:58:18 mr_checkout.go:36: the key `remote.origin.url` is not found
$ lab mr checkout origin-2 2
fatal: 'origin' does not appear to be a git repository
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
2020/09/25 17:58:24 mr_checkout.go:79: exit status 128
```
IIUC (and, please, correct me if I'm wrong) these commands rely on the information held in the `forkedFromRemote` (and `forkRemote` (?)) internal vars, which are always set to one of the options: `origin` or `upstream` (and `lab.User()` for `forkRemote` ), which doesn't seem to be the best approach, considering the users may choose different naming convention.

A way of solving it would be to rework the way `forkedFromRemote` gets set, by also considering/checking the current remote being tracked by the current branch before settling on `origin`. Another point I need some lights on is on the usage of `forkRemote`, that I didn't really understand.

zaquestion:
Bit of a brain dump... bare with me.

The primary purpose of `forkRemote` and `forkFromRemote` are to add the needed context for a few different types of scenarios, mostly around choosing the right default remote for a command:
1. Creating merge requests, in the absence of the branch being tracked, lab can make the assumption that you want to create a merge request from your fork to the original. Might be some other cases like this too
[...]

PR #440, opened 2020-09-25
add fallthrough path for forkedFromRemote

bmeneguele:
Mostly related to issue #438, this patchset fix the discovery of the remote the user forked from and also adds a fallthrough path in case none of the default options are met.

A test code was added to the new internal/git/git.go:GetLocalRemotesFromFile() function: although not entirely necessary we need to make sure the `git config --get-regex` behavior doesn't change over time.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -253,9 +253,9 @@ func setCommandPrefix(scmd *cobra.Command) {
 
 var (
 	// Will be updated to upstream in Execute() if "upstream" remote exists
-	forkedFromRemote = "origin"
+	forkedFromRemote = ""
 	// Will be updated to lab.User() in Execute() if forkedFrom is "origin"
-	forkRemote = "origin"
+	forkRemote = ""
 )
 
 // Execute adds all child commands to the root command and sets flags appropriately.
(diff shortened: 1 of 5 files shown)


cmd/root.go
```diff
@@ -265,21 +265,39 @@ func Execute() {
...
 	}
+	_, err = gitconfig.Local("remote.origin.url")
+	if err == nil {
+		forkedFromRemote = "origin"
+	}
+
 	if forkedFromRemote == "" {
-		// use the remote tracked by the branch if set
+		// use the remote tracked by the default branch if set
 		masterRemote, err := gitconfig.Local("branch.master.remote")
```
zaquestion, line 275: Probably worthwhile with the big shift from `master` to `main`, to include `branch.main.remote` as a case in here as well.

bmeneguele: Just pushed a new version of the patch.

bmeneguele:
Ok, it seems there are other tests I need to check the results.

zaquestion:
This is looking good so far, just a matter of sorting out the tests.

bmeneguele:
I added a small fix to a non-related test case on "issue_show", just for completeness.

cmd/root_test.go
```diff
@@ -69,6 +69,9 @@ func TestMain(m *testing.M) {
 	}
 	lab.Init(host, u.Username, token, false)
 
+	// Make "origin" the default remote for testing
+	forkedFromRemote = "origin"
+	forkRemote = "origin"
```
zaquestion, line 74: This seems like it would mask failure cases. `origin` is the first remote listed in https://github.com/zaquestion/lab/blob/master/testdata/test.git/config which will be used in any tests that run with `copyTestRepo()`

Although this line seems to be a copypaste error, the branch in this case should be `master` https://github.com/zaquestion/lab/blob/master/testdata/test.git/config#L14 

Still I'd expect the logic you have above to detect `origin` as the correct remote to use, by seeing it as the first remote in this list.

zaquestion: Oh actually I guess these lines would only apply to tests that _don't_ run using the test binary. So this is probably fine

bmeneguele: The issue is that the tests that failed are those that run the cmd.Run() directly, without calling `exec.Command(labBinaryPath, args...).`, with that, the `root.go:Execute()` is just bypassed and both `forkFromRemote` and `forkRemote` are left empty. Before the patch both vars were getting set to `origin` because they were being assigned to that in compile time. I should re-word the patch description to point to what I just have said.

bmeneguele: > Oh actually I guess these lines would only apply to tests that don't run using the test binary. So this is probably fine

Yes, exactly :). I didn't notice your message before I sent mine.
I'll re-word the commit log to point it.

zaquestion: Sorry to make you do another thing, but can you expand the [code] comment to indicate that as well? After that I think this is ready to come in

bmeneguele: No worries at all :). I just pushed the new patch with updated comment.

zaquestion (approved):

zaquestion:
Found a small bug when working on something else https://github.com/zaquestion/lab/issues/446

PR #441, opened 2020-09-26
fix: (mr checkout) override target remote when passed as arg

zaquestion:
despite accepting the remote as an arg and looking up the MRs on the correct project, fetchs were still being done on the forkedFromRemote when they should also have been carried out on the override


bmeneguele (approved):
Yep, it solves the issue with `mr checkout` that I pointed in issue #438.
Thanks @zaquestion !

PR #458, opened 2020-10-02
root: skip test instead of fail if running locally

bmeneguele:
Test_config_gitConfig_* tests were failing while I was trying to run `go test` locally.
The reason was that these tests are looking for `/home/travis/...` path, which, obviously doesn't exist.
However, at the same time, I also understand why the tests don't use a "generic" home dir: it would erase/modify my local lab config.
Because of that, I thought that instead of failing, the tests could only be skipped in case the path is invalid.

### DIFF
--- a/cmd/root_test.go
+++ b/cmd/root_test.go
@@ -518,10 +518,11 @@ func getAppOutput(output []byte) []string {
 	return lines
 }
 
-func setConfigValues(repo string, configVal string, gitVal string) {
+func setConfigValues(repo string, configVal string, gitVal string) error {
 	err := os.Rename(repo+"/lab.toml", "/home/travis/.config/lab/lab.toml")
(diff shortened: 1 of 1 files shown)


zaquestion (commented):
Not sure how I feel about bringing this in. In general having paths tied to travis isn't great, especially as we're moving to GitLab CI. Would like to find a more portable approach. I fear if we add these skips we'll forget about these tests when we're off travis.

bmeneguele:
Although they're being skipped their error messages are pretty loudy. Once a while we have to add new tests and we may end up seeing the errors there. I understand your concern, but not being able to run tests locally without forcing the developer to manually skip/comment the test cases is somewhat frustrating (at least to me hehehe).

zaquestion (approved):

PR #463, opened 2020-10-13
config: Add a better user authentication error message

prarit:
The user authentication can fail if the token or load_token field is
incorrect.  lab fails with, for example,

2020/10/13 13:29:46 config.go:194: GET https://gitlab.com/api/v4/user: 401 {message: 401 Unauthorized}

which is not human-readable.

Output a better and human friendly authentication error message.


prarit:
I'm going to add a test for this before merging.  Should be simple to add something to cmd/root_test.go that adds a borken PAT and then executes lab.

Issue #466, opened 2020-10-14
Add pushRemote and pushDefault for source remote


bmeneguele:
git-config(1) mentions two options that 'lab' could use for determining the source remote of a certain repository in addition to (1) `branch.<name>.remote`, they are: (2) `remote.pushDefault` and (3) `branch.<name>.pushRemote`. These three options obay the following precedence order: (3) > (2) > (1).

Enabling these options help those users that don't have a tracking remote for every branch, but use, for instance, the `remote.pushDefault` instead.

bmeneguele:
Considering MR #467 was already merged we can close it.

PR #467, opened 2020-10-14
mr_create: add git-config options as possible source remotes

bmeneguele:
Solves issue #466 by adding `remote.pushDefault` and `branch.<name>.pushRemote` as possible source remote options.
Also, the patch also adds a test case for the function.


zaquestion (approved):

zaquestion:
Those test failures might be an indicator of something changing on GitLab's side.

PR #470, opened 2020-10-15
ci_status_test: Reverse expected output

prarit:
The Travis CI results are now reversed and are causing ci_status_test
to fail when checking the expected output.

Reverse the excepted output.


zaquestion:
There is a slight functionality break/change here. The case "noarg" case by design is intended to show logs ("trace") for the most recently run (or active job), which is achieved by the ordering of the jobs list IIRC. I should have commented this behavior more clearly in the test case. 

BTW, do we have any idea if this resulted from a change on our side or GitLabs? If possible I try to find a GL issue/PR to reference when changes happen.

zaquestion:
All that said, not to detract from the quick fix! Thanks @prarit !

PR #477, opened 2020-10-30
Make http default fork

eedgar:


bmeneguele:
Hi @eedgar . It would be necessary to change some testing code in order to get this MR accepted for merge.
Also, I think that only changing the default option without keeping an option for those using SSH is not the correct approach either. I'm not against of changing the default to HTTPS (I don't have any information on what's the most used protocol), but at least an "--ssh" should be available since it has been the default for quite some time.

Would you mind working on it? Or do you prefer someone else come up with this options?

Thanks!

bmeneguele:
@eedgar , I'm going to close this MR in favor of #496, which handles other commands other than `fork`.
Please, if possible, test the code changes I did there, to make sure your workflow runs correctly.
Many thanks!

PR #486, opened 2020-11-09
edit: ignore empty argument

bmeneguele:
Golang exec.Command() considers empty args as valid/normal arguments to be
passed to the program to operate against.  In the case of a text editor an
empty filepath may causes different actions depending on the one the user
uses, for example: (1) a "new buffer" (nano) or (2) the current path (vim)
are opened for edit.

With that, when the user call "lab mr create" without specifying a file (-F)
the text editor open two files for edit: the "empty file" (triggering an
undefined behavior) and the actual ISSUE_NOTE/MR_NOTE/MERGEREQ.  This patch
adds a conditional to ignore the args passed to the text editor in case it's
empty, avoiding such behavior.

### DIFF
--- a/internal/git/edit.go
+++ b/internal/git/edit.go
@@ -83,8 +83,12 @@ func editorCMD(editorPath, filePath string) *exec.Cmd {
 		args = append(args, "--cmd", "set ft=gitcommit tw=0 wrap lbr")
 	}
 	argparts := strings.Join(parts[1:], " ")
-	argparts = strings.Replace(argparts, "'", "\"", -1)
-	args = append(args, argparts, filePath)
+	if len(argparts) == 0 {
+		args = append(args, filePath)
+	} else {
+		argparts = strings.Replace(argparts, "'", "\"", -1)
+		args = append(args, argparts, filePath)
+	}
 	cmd := exec.Command(parts[0], args...)
 	cmd.Stdin = os.Stdin
 	cmd.Stdout = os.Stdout


prarit:
With that, when the user call "lab mr create" without specifying a file (-F)
the text editor open two files for edit: the "empty file" (triggering an
undefined behavior) and the actual ISSUE_NOTE/MR_NOTE/MERGEREQ.  This patch
adds a conditional to ignore the args passed to the text editor in case it's
empty, avoiding such behavior.

@bmeneguele -- I just tried this:

'lab mr create'

and

'lab mr create -F' 

and cannot reproduce this behaviour.  Can you give me an example of the failing command?

bmeneguele:
@prarit if you have a GIT_EDITOR (`git config core.editor`) with some appended args it won't fail, but with a pure and simple `nano` or `vim` it should happen. This issue is annoying me for quite some time already, I just noticed the real issue when I, by curiosity, changed the GIT_EDITOR to `nano` and noticed the "NEW BUFFER" written where it should've been the filepath.

bmeneguele:
@prarit, for instance:
```
$ git config core.editor
nano
$ lab mr create
```
nano-first

And right after I press `ctrl+x` to close the file:

nano-second

The same happens using `vim` (or `nvim`), the difference is that it opens the current folder for edit:

nvim-first

Unless I have something configured in a really non-standard way, it happens 100% the time and the patch resolves it.

prarit:
I wish there was a way to test this. It's been broken a few times now, ie) 3f4e7ba4f6ff ("edit: Fix GIT_EDITOR evaluation").  In any case, WFM.

prarit (approved):
LGTM.

PR #492, opened 2020-11-14
config: check mandatory config opts before moving on

bmeneguele:
It seems to be a good idea to first check the mandatory config opts before leaving/returning from `LoadMainConfig`, avoiding strange error messages as `lab` tries to perform actions with empty opts.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -309,8 +309,16 @@ func LoadMainConfig() (string, string, string, string, bool) {
 		}
 	}
 
-	host = MainConfig.GetString("core.host")
-	token = GetToken()
+	if !MainConfig.IsSet("core.host") {
+		host = defaultGitLabHost
+	} else {
+		host = MainConfig.GetString("core.host")
+	}
+
+	if token = GetToken(); token == "" {
+		UserConfigError()
+	}
+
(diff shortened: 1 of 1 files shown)


prarit:
LGTM -- OOC what commands failed?

bmeneguele:
In case the `core.host` was missing but with the 
```
$ lab mr list
2020/11/17 10:45:50 mr_list.go:37: Get "/api/v4/projects/Test-Lab": unsupported protocol scheme ""
Error: User authentication failed.  This is likely due to a misconfigured Personal Access Token.  Verify the token or token_load config settings before attemping to authenticate.
```

And in case `core.token` as missing:
```
$ lab mr list
2020/11/17 10:48:21 mr_list.go:37: Get "/api/v4/projects/bmeneg%2FTest-Lab": unsupported protocol scheme ""
Error: User authentication failed.  This is likely due to a misconfigured Personal Access Token.  Verify the token or token_load config settings before attemping to authenticate.
```

The `Error: ` line is correct, but the `log.Fatal()` ones should not happen: it should fail earlier IMO. That's why I added the conditionals to `config.go`.

prarit (approved):
Approved.

PR #495, opened 2020-11-17
config: home dir is only useful if XDG_CONFIG_HOME isn't set

bmeneguele:
Instead of setting the 'home' var early, only set it in case XDG_CONFIG_HOME
isn't set. 'home' var is only useful for composing the config path, nothing
else.


bmeneguele:
I'm merging this MR without a review because it's a trivial change with no code behavior change.

Issue #345, opened 2019-10-22
lab fork should allow to specify the protocol


mfriedenhagen:
* Right now `lab` seems to default to using `ssh` when forking stuff.
* Using `ssh` is not possible in our internal instance.
* `lab` uses an API token anyways to do stuff, so the only "known, good way" to reach the GitLab instance is https anyways.

My suggestion would be to default to use `https` and make `ssh` an option.

eedgar:
We also have the same issue.

eedgar:
https://github.com/zaquestion/lab/pull/477

bmeneguele:
It's now solved.
However, the default protocol still is SSH. For using HTTP call the command with `--http`.

PR #496, opened 2020-11-18
Enable HTTP protocol for different commands

bmeneguele:
This MR bring the patch created by @eedgar in MR #477, which turns HTTP protocol as the default for forking a repo.
However, it turns SSH off and let other places that relies on the protocol format without changes.
The additional patch(es) make HTTP possible through a command flag (`--http`), letting SSH still as the default.
But now, the user can `lab fork --http`, `lab clone --http`, and others, to use the HTTP URL scheme instead of the SSH.
Quick note for reviewers: the GitLab API token for HTTP need to have both `{read,write}_repository` capabilities set since they're the ones allowing GIT-over-HTTP trafic.

### DIFF
--- a/cmd/clone.go
+++ b/cmd/clone.go
@@ -32,7 +32,7 @@ var cloneCmd = &cobra.Command{
 		} else if err != nil {
 			log.Fatal(err)
 		}
-		path := project.SSHURLToRepo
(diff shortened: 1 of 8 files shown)


bmeneguele:
This MR also solves the issue #345

PR #497, opened 2020-11-19
lab: Move issues from one project to another

prarit:
In large projects with groups, and extended namespaces, issues
occasionally get filed in the wrong project.  GitLab's API provides
functionality to move an issue from one project to another.

Add 'lab issue move <new_project>'.

### DIFF
--- a/cmd/issue_move.go
+++ b/cmd/issue_move.go
@@ -1,0 +1,51 @@
+package cmd
+
+import (
+	"fmt"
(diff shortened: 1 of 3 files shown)


prarit:
@zaquestion , there isn't a good way to test this functionality.  Perhaps adding a new test repo (test2?) to gitlab so that issues can be moved back and forth would be a solution?

PR #511, opened 2020-12-03
mr: Don't set minimum number of args

fmuellner:
The commands use parseArgsWithGitBranchMR() to parse non-flag arguments,
which will return the MR id associated with the current branch if no
id is specified.

That behavior makes a lot of sense, but can only work if the option parser
doesn't enforce the id to be specified.


prarit:
LGTM.

prarit (approved):

bmeneguele (approved):

PR #512, opened 2020-12-04
mr/show: Fix glitch in error message

fmuellner:
The remote name in the output isn't separated from the next word.


prarit:
@fmuellner I kicked your build again.  The error is a weird occasional error and re-kicking the build seems to resolve it.

PR #513, opened 2020-12-04
mr: Remove questionable aliases

fmuellner:
Those are likely copy+paste errors, in particular merging is pretty
much the opposite of deleting ...


prarit (approved):
I feel like I found a few other aliases like this a while ago.  LGTM, and thanks.

Issue #471, opened 2020-10-26
fix(ci): view is horizontally reversed


davinkevin:
Following a recent upgrade of GitLab (13.5 I think), the result of the `lab ci view` command is totally reversed, on **right to left** axis. 

![image](https://user-images.githubusercontent.com/1970922/97146877-e58a0f00-1768-11eb-9b18-4bfa51923c2f.png)

![image](https://user-images.githubusercontent.com/1970922/97146927-f8044880-1768-11eb-9006-3a313d40a287.png)

Last time I use it was on friday, and today all is **right to left**. 

I didn't change anything to my configuration (I didn't turn the computer on during the week end), and this happen in two different GitLab Instances (gitlab.com and a private managed one).

/cc @neonox31

zaquestion:
@davinkevin i believe this is fixed in main, i remember a test breaking... ill prioritize getting a release out after halloween, im moving atm and my whole life is packed up 😅

davinkevin:
No problem, it's a bit strange to have this behavior but I will wait until you unpack everything.

BTW, moving IRL and from GitHub to GitLab in parallel is funny 😇

galan:
Also affected. Waiting for release since https://github.com/zaquestion/lab/issues/410 . None of the checkboxes is checked since then on https://github.com/zaquestion/lab/issues/406 .

Would be creating a last intermediate release an option to fix this and have the latest changes available (that would fit your packed timetable)? Thanks for the great tool.

davinkevin:
For your information @galan , this is pretty simple to build a version for your OS. Checkout the code, modify the `MAKEFILE` to include your os/arch and `make install`. 

I've tested it, but the `ci view` is still reversed and it totally crash after launching this command 😓. I'll go back to the previous version…

davinkevin:
BTW, there is still some problem with job relauched… they end up at the end of the graph and don't keep the same position. 

Would you like me to open another issue ? 

/cc @fmuellner && @zaquestion

fmuellner:
I've seen that as well - both before and after the fix, so it's a separate issue.

Not sure there's an easy fix though, I didn't find any API for getting stages in order (or at all), only through jobs.

PR #514, opened 2020-12-06
ci: Sort list of displayed jobs

fmuellner:
When we filter the list to only show the latest job of each stage+name
combo, we mess up the expected order of jobs in the pipeline.

Address this by sorting the filtered list before returning it.

Fixes https://github.com/zaquestion/lab/issues/471

### DIFF
--- a/cmd/ci_status_test.go
+++ b/cmd/ci_status_test.go
@@ -41,25 +41,25 @@ func Test_ciStatus(t *testing.T) {
 	}
 	out := string(b)
 	assert.Contains(t, out, `Stage:  Name                           - Status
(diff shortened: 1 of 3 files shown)


bmeneguele (approved):

bmeneguele (changes requested):
Actually, this change seems to be breaking the order of expected test results.
Could you take a look on both `Test_latestJobs` and `Test_ciStatus`?

fmuellner:
Updated:
 - code moved to CIJobs(), which on second thought seems a better place for sorting
 - reverted tests to their original order

bmeneguele (approved):

PR #515, opened 2020-12-06
Use the remote tracking branch where appropriate

### LINKED ISSUE #360: `lab ci status` when local branch name != remote branch name

fmuellner:
While the local branch name usually matches the remote one, this is
not necessarily the case, and the automatic detection of a MR or pipeline
fails.

Handle those cases as well by figuring out the remote tracking branch
and use that instead of the local name if possible.

Fixes #360

### DIFF
--- a/cmd/ci_status_test.go
+++ b/cmd/ci_status_test.go
@@ -17,14 +17,14 @@ func Test_ciStatus(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	cmd = exec.Command(labBinaryPath, "checkout", "origin/ci_test_pipeline")
+	cmd = exec.Command(labBinaryPath, "checkout", "-b", "ci_test_pipeline")
 	cmd.Dir = repo
 	if b, err := cmd.CombinedOutput(); err != nil {
 		t.Log(string(b))
 		t.Fatal(err)
 	}
 
-	cmd = exec.Command(labBinaryPath, "checkout", "-b", "ci_test_pipeline")
+	cmd = exec.Command(labBinaryPath, "branch", "-m", "local/ci_test_pipeline")
 	cmd.Dir = repo
 	if b, err := cmd.CombinedOutput(); err != nil {
 		t.Log(string(b))
--- a/cmd/mr_test.go
+++ b/cmd/mr_test.go
@@ -109,7 +109,7 @@ func Test_mrCmd_MR_description_and_options(t *testing.T) {
 		}
 	})
 	t.Run("create MR from file", func(t *testing.T) {
-		git := exec.Command("git", "checkout", "mrtest")
+		git := exec.Command("git", "checkout", "-b", "local/mrtest", "origin/mrtest")
 		git.Dir = repo
 		b, err := git.CombinedOutput()
 		if err != nil {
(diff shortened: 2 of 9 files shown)


bmeneguele (approved):
Changes look good to me.
However, I think that having a test for this behavior, to avoid messing with it in the future is worthy.
It can be sent as another MR if desired.

fmuellner:
> However, I think that having a test for this behavior, to avoid messing with it in the future is worthy.

OK, this took a bit :-)

Was that more or less what you had in mind?

cmd/mr_test.go
```diff
@@ -109,7 +109,7 @@ func Test_mrCmd_MR_description_and_options(t *testing.T) {
 		}
 	})
 	t.Run("create MR from file", func(t *testing.T) {
-		git := exec.Command("git", "checkout", "mrtest")
+		git := exec.Command("git", "checkout", "-b", "local/mrtest", "origin/mrtest")
```
prarit, line 112: I'm a bit wary of overloading a the "create MR from file" test this way.  Instead of this I suggest adding a completely new test for this functionality.

cmd/ci_status_test.go
```diff
@@ -17,14 +17,14 @@ func Test_ciStatus(t *testing.T) {
...
-	cmd = exec.Command(labBinaryPath, "checkout", "origin/ci_test_pipeline")
+	cmd = exec.Command(labBinaryPath, "checkout", "-b", "ci_test_pipeline")
 	cmd.Dir = repo
 	if b, err := cmd.CombinedOutput(); err != nil {
 		t.Log(string(b))
 		t.Fatal(err)
 	}
 
-	cmd = exec.Command(labBinaryPath, "checkout", "-b", "ci_test_pipeline")
+	cmd = exec.Command(labBinaryPath, "branch", "-m", "local/ci_test_pipeline")
```
prarit, line 27: Change this to be 'git' not 'labBinaryPath'. (please :))

bmeneguele:
Yep! :)
Thanks for the changes!

prarit:
Looks like @bmeneguele & I crossed streams :)  @fmuellner, any chance I could ask for a follow on changeset to fix the issues I raised?

fmuellner:
> @fmuellner, any chance I could ask for a follow on changeset to fix the issues I raised?

Mmh, where was that? I'm happy to address any issues, but I'm not seeing any comments here.

bmeneguele:
> Looks like @bmeneguele & I crossed streams :) @fmuellner, any chance I could ask for a follow on changeset to fix the issues I raised?

Ouch! What a timing hm? :-/
Unfortunately, I can't see the issue you raised, wasn't it closed by my merge?

prarit:
Ouch and I even forgot to click the "Add Review Comments" button to completely screw up everything.  In any case ... some simple review comments.

PR #522, opened 2020-12-10
mr_reopen: add command for reopening a closed MR

bmeneguele:
Add `lab mr reopen` command.

### DIFF
--- a/cmd/mr_list_test.go
+++ b/cmd/mr_list_test.go
@@ -123,9 +123,7 @@ func Test_mrFilterByTargetBranch(t *testing.T) {
 	assert.Empty(t, mrs, "Expected to find no MRs for non-existent branch")
 }
 
-var (
-	latestUpdatedTestMR = "#18 MR for approvals and unapprovals"
-)
+var latestUpdatedTestMR = "#19 MR for closing and reopening"
 
 func Test_mrListByTargetBranch(t *testing.T) {
 	t.Parallel()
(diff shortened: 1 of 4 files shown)


prarit:
I like this, but would creating the files as mr_open.go and mr_open_test.go, instead of mr_reopen.go and mr_reopen_test.go be better?  You could add a command alias for 'reopen'.  It would then balance nicely with the mr_close files.

prarit:
@bmeneguele I've taken your latest MR test fix and am going to add it separately in #523 .  FYI.

bmeneguele:
> I like this, but would creating the files as mr_open.go and mr_open_test.go, instead of mr_reopen.go and mr_reopen_test.go be better? You could add a command alias for 'reopen'. It would then balance nicely with the mr_close files.

Hmmm, that's an option indeed. However, won't it create a "confusion" between `lab mr create` and `lab mr open` ? Although they do completely different things, the words are quite close to their meaning. OTOH `reopen` has a clear and obvious behavior: it won't create nor open anything, it'll only reopen a closed MR.

prarit:
> > I like this, but would creating the files as mr_open.go and mr_open_test.go, instead of mr_reopen.go and mr_reopen_test.go be better? You could add a command alias for 'reopen'. It would then balance nicely with the mr_close files.
> 
> Hmmm, that's an option indeed. However, won't it create a "confusion" between `lab mr create` and `lab mr open` ? Although they do completely different things, the words are quite close to their meaning. OTOH `reopen` has a clear and obvious behavior: it won't create nor open anything, it'll only reopen a closed MR.

That's a good counter point :).  There's no good answer here.  Stick with "reopen" ;)

prarit (approved):

PR #524, opened 2020-12-11
fork: enable different name, namespace and path on destination

### LINKED ISSUE #488: fork: allow user to set name and path
When the fork is issued from the API it's possible to the user to set a different name (to avoid conflict) and also a different path (besides the user's own namespace) to fork the project.

It's useful for workflows that inside a group each user has it's own namespace, e.g.: 
`gitlab.com/company/devops/users/joe`
`gitlab.com/company/devel/users/brian`
...
while they still have their own personal namespace:
`gitlab.com/joe`
`gitlab.com/brian`

Currently, `lab` only supports the personal namespace path, which is automatically chosen.
IMHO it would be nice to allow the user to specify a different name and path in case he wishes, while keeping the personal namespace as the default approach.

API reference: https://docs.gitlab.com/ee/api/projects.html#fork-project

bmeneguele:
Allow the user to specify three fork properties: `name`, `namespace` and `path`.
With it the user can fork the same project to different namespaces with different names, working around possible conflicts.

Fixes #488

### DIFF
--- a/cmd/fork.go
+++ b/cmd/fork.go
@@ -11,6 +11,7 @@ import (
 )
 
 var skipClone = false
+var forkData lab.ForkStruct
 
 // forkCmd represents the fork command
 var forkCmd = &cobra.Command{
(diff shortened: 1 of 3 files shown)


bmeneguele:
I still have to create the tests for the new code. Because of that I'm going to set this MR as draft.

bmeneguele:
@prarit do you think you could take a look at the changes? Just to make sure I've met the expectations.
I wasn't able to fully test the changes, even though I created a testing group in the lab-testing user.
I plan to improve it as we go.

PR #526, opened 2020-12-13
Add support for draft merge requests

fmuellner:
Gitlab supports marking a merge request as draft to indicate that it is not ready to be merged. This is supported in the web UI (actions to mark as draft/ready, disabled merge button in draft MRs), but not explicitly exposed in the API: Drafts are simply indicated by prefixing the MR title with "Draft:", "[Draft]" or "(Draft)" (or previously "WIP)[0].

lab therefore kind of supports the feature already, but it still seems reasonable to export the feature explicitly rather than via manipulating the title (although that's of course what the implementation does under the hood).

[0] https://docs.gitlab.com/ee/user/project/merge_requests/work_in_progress_merge_requests.html

### DIFF
--- a/cmd/mr_list.go
+++ b/cmd/mr_list.go
@@ -19,6 +19,8 @@ var (
 	mrNumRet       int
 	mrAll          bool
 	mrMine         bool
+	mrDraft        bool
+	mrReady        bool
 	assigneeID     *int
 	mrAssignee     string
 	order          string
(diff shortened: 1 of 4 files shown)


cmd/mr_list.go
```diff
@@ -105,6 +115,8 @@ func init() {
 		&mrAssignee, "assignee", "", "list only MRs assigned to $username")
 	listCmd.Flags().StringVar(&order, "order", "updated_at", "display order, updated_at(default) or created_at")
 	listCmd.Flags().StringVar(&sortedBy, "sort", "desc", "sort order, desc(default) or asc")
+	listCmd.Flags().BoolVarP(&mrDraft, "draft", "", false, "list MRs marked as draft")
+	listCmd.Flags().BoolVarP(&mrReady, "ready", "", false, "list MRs not marked as draft")
```
bmeneguele, line 120: What if we don't use 2 flags here, but have only `--draft` to denote we want to also list draft MRs, while calling `lab mr list` with no flags always list only ready MRs?

bmeneguele: Forget it.. makes no sense at all.

bmeneguele (approved):

PR #527, opened 2020-12-14
Add support for merge request pipelines

fmuellner:
Gitlab allows setting up pipelines that are associated with a
merge request rather than a branch/tag ("detached")[0].

Merge request pipelines may be created instead of or in addition
to the regular branch/tag pipeline, and can use a different set
of jobs.

When merge request pipelines are used, they control whether a MR
can be merged, so their status is clearly interesting.

Unfortunately if both branch and MR pipelines are enabled, which
one ends up being the last one associated with a commit is random.

Address this by adding an explicit --merge-request flag to the
various ci subcommands to explicitly refer to the detached pipeline.

[0] https://docs.gitlab.com/ee/ci/merge_request_pipelines/#pipelines-for-merge-requests

### DIFF
--- a/cmd/ci_status.go
+++ b/cmd/ci_status.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"log"
 	"os"
+	"strings"
 	"text/tabwriter"
 	"time"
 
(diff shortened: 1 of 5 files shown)


bmeneguele (approved):
LGTM.

PR #528, opened 2020-12-14
RFC: Use bang to denote MR IDs

fmuellner:
In Gitlab's UI, issues appear as #123 and merge requests as !123.
Follow that convention when outputting MR IDs.


bmeneguele (approved):
Although `lab` allow passthrough to `hub` I agree it should reflect GitLab's behavior.
So +1 for this change from me.

PR #533, opened 2020-12-19
Support duplicating and reopening issues

fmuellner:
Right now there's support for reopening merge requests, but not issues.

Plus when closing an issue, it can be useful to mark it as a duplicate of another issue.

### DIFF
--- a/internal/gitlab/gitlab.go
+++ b/internal/gitlab/gitlab.go
@@ -604,6 +604,25 @@ func IssueClose(pid interface{}, id int) error {
 	return nil
 }
 
+// IssueDuplicate closes an issue as duplicate of another
+func IssueDuplicate(pid interface{}, id int, dupId string) error {
+	// Not exposed in API, go through quick action
+	body := "/duplicate " + dupId
+
+	_, _, err := lab.Notes.CreateIssueNote(pid, id, &gitlab.CreateIssueNoteOptions{
+		Body: &body,
+	})
+	if err != nil {
+		return errors.Errorf("Failed to close issue #%d as duplicate of %s", id, dupId)
+	}
+
+	issue, _, err := lab.Issues.GetIssue(pid, id)
+	if issue == nil || issue.State != "closed" {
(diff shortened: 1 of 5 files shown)


bmeneguele (approved):

internal/gitlab/gitlab.go
```diff
@@ -604,6 +604,25 @@ func IssueClose(pid interface{}, id int) error {
 	return nil
 }
 
+// IssueDuplicate closes an issue as duplicate of another
+func IssueDuplicate(pid interface{}, id int, dupId string) error {
+	// Not exposed in API, go through quick action
+	body := "/duplicate " + dupId
```
bmeneguele, line 610: I wasn't able to find about that in the API docs too.. but seems a good enough solution to use the 'quick action' mechanism :-)

fmuellner: I also looked at the raw json output for an issue that was closed as duplicate, and the issue it duplicated. There's really nothing there :shrug:

fmuellner:
By the way, is there some trick for fixing the coverage test? If I read https://codecov.io/gh/zaquestion/lab/pull/533/diff correctly, the code that calls `IssueDuplicate`/`IssueReopen` is called, but not a single line in those functions is covered ... which is rather baffling?!

bmeneguele:
TBH, that's a good question. I've faced the same issue other times, but never really looked into for details.
@zaquestion had plans to rework the coverity IIRC, but I don't think we have a workaround for it right now.
So I think it's fine just ignoring, for now.

prarit:
> By the way, is there some trick for fixing the coverage test? If I read https://codecov.io/gh/zaquestion/lab/pull/533/diff correctly, the code that calls `IssueDuplicate`/`IssueReopen` is called, but not a single line in those functions is covered ... which is rather baffling?!

I've noticed that before and @zaquestion has too.  It's annoying and I haven't figured out a reason why codecov makes this error on occasion.  In any case, you've added a test and that's really what is important :)

PR #534, opened 2020-12-20
util: Always override config file flags from the command line

fmuellner:
Currently we check whether a flag's current value differs from its
default in order to skip the value from the config file. That works
in most cases, but as a result it is not possible to override a value
from the config by explicitly passing the default value:

  [mr_show]
  comments = true

  $ lab mr show --comments=false

Fix this by checking the flag's Changed property instead.

### DIFF
--- a/cmd/util.go
+++ b/cmd/util.go
@@ -56,7 +56,7 @@ func flagConfig(fs *flag.FlagSet) {
 		}
 
 		// if set, always use the command line option (flag) value
-		if f.Value.String() != f.DefValue {
+		if f.Changed {
 			return
 		}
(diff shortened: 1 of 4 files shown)


cmd/util.go
```diff
@@ -56,7 +56,7 @@ func flagConfig(fs *flag.FlagSet) {
 		}
 
 		// if set, always use the command line option (flag) value
-		if f.Value.String() != f.DefValue {
+		if f.Changed {
```
prarit, line 59: Interesting.  I had no idea f.Changed was an option :)

prarit (approved):

PR #536, opened 2020-12-20
Disable flag sorting for some commands

fmuellner:
By default, flags are sorted alphabetically. However there are some
cases where it makes more sense to group related flags together,
like --label/--unlabel, --assign/--unassign or --draft/--ready.

Disable flag sorting in those cases.


bmeneguele:
I'm fine with it :)

bmeneguele (approved):

PR #537, opened 2020-12-20
ci_artifacts: Allow downloading artifact(s)

### LINKED ISSUE #232: Support download of CI-uploaded artifacts
It would be really helpful to be able to download artifacts uploaded during CI process.
In Gitlab the artifacts can be downloaded through `/jobs/{jobId}/artifacts/download`. 

Example of useful command to download the artifacts of the CI build:
`lab ci download`
or: 
`lab ci artifacts`

fmuellner:
Quoting the commit message:

    Pipeline jobs can produce files that are uploaded to gitlab and made
    available for download.

    Support this with a new `ci artifacts` subcommand, which either
    downloads the entire artifacts archive, or a singe file from
    that archive when the --artifact-path option is used.

I originally implemented the single-artifact case as positional argument
as discussed in #232, i.e. "remote branch:job path", but then decided
against it:

 - I like how `lab` generally does the right thing when a positional
   argument is omitted, but that's tricky when there are four of them;
   in particular job name and path are problematic, as those are just
   arbitrary strings as far as we know

 - command line flags can be set up in the config file (for example
   if the pipeline only produces a single "interesting" artifact)

The new test case uses the new job added in
https://gitlab.com/zaquestion/test/-/merge_requests/22,
so for now it's expected to fail.

Fixes #232

### DIFF
--- a/cmd/ci_artifacts.go
+++ b/cmd/ci_artifacts.go
@@ -1,0 +1,79 @@
+package cmd
+
+import (
+	"fmt"
+	"io"
+	"log"
+	"os"
+
+	"github.com/rsteube/carapace"
+	"github.com/spf13/cobra"
+	"github.com/zaquestion/lab/internal/action"
+	lab "github.com/zaquestion/lab/internal/gitlab"
+)
+
+var ciArtifactsCmd = &cobra.Command{
+	Use:              "artifacts [remote [[branch:]job]]",
+	Short:            "Download artifacts of a ci job",
+	Long:             `If a job is not specified the latest job with artifacts is used`,
+	PersistentPreRun: LabPersistentPreRun,
+	Run: func(cmd *cobra.Command, args []string) {
+		var (
+			rn      string
+			jobName string
+			err     error
+		)
+		jobName, branchArgs, err := filterJobArg(args)
+		if err != nil {
+			log.Fatal(err)
+		}
+
+		forMR, err := cmd.Flags().GetBool("merge-request")
+		if err != nil {
+			log.Fatal(err)
+		}
+
+		path, err := cmd.Flags().GetString("artifact-path")
+		if err != nil {
+			log.Fatal(err)
+		}
+
+		rn, pipelineID, err := getPipelineFromArgs(branchArgs, forMR)
+		if err != nil {
+			log.Fatal(err)
+		}
+
+		project, err := lab.FindProject(rn)
+		if err != nil {
+			log.Fatal(err)
+		}
+		projectID := project.ID
+
+		r, outpath, err := lab.CIArtifacts(projectID, pipelineID, jobName, path)
+		if err != nil {
+			log.Fatal(err)
+		}
+
+		dst, err := os.Create(outpath)
+		if err != nil {
+			log.Fatal(err)
+		}
+
+		_, err = io.Copy(dst, r)
+		if err != nil {
+			log.Fatal(err)
+		}
+
+		fmt.Printf("Downloaded %s\n", outpath)
+	},
+}
+
+func init() {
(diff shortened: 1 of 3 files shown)


bmeneguele:
That's something I was actually looking at to implement :))
Thanks for that! Even though the code seems simple and correct I'm going to take it a try... I should have feedback pretty soon.

fmuellner:
> The new test case uses the new job added in
> https://gitlab.com/zaquestion/test/-/merge_requests/22,
> so for now it's expected to fail.

Mmh, I managed to break all tests ... again :frowning_face: 

Should I better close it?

> Thanks for that! Even though the code seems simple and correct I'm going to take it a try... I should have feedback pretty soon.

Sure, no hurry. And thanks!

bmeneguele:
> > The new test case uses the new job added in
> > https://gitlab.com/zaquestion/test/-/merge_requests/22,
> > so for now it's expected to fail.
> 
> Mmh, I managed to break all tests ... again frowning_face
> 
> Should I better close it?
> 

Not really. You can turn this PR into a "draft" and keep force pushing the changes until it's solved.

bmeneguele:
@fmuellner I'm going to set this PR as draft, just to avoid merging it by accident :). But feel free to remote the draft status after you have the changes done.

bmeneguele:
@fmuellner same thing here, I'm not being able to re-run the tests =(
But at the same time it still seems to be failing with specific ci_artifacts code.

fmuellner:
Yeah, the new tests assume that https://gitlab.com/zaquestion/test/-/merge_requests/22 is merged. I can look into some alternative if necessary.

bmeneguele:
> Yeah, the new tests assume that https://gitlab.com/zaquestion/test/-/merge_requests/22 is merged. I can look into some alternative if necessary.

It's now merged. Could you retry?

fmuellner:
> Could you retry?

It'll need #557 first ...

bmeneguele:
> > Could you retry?
> 
> It'll need #557 first ...

Oh, indeed! Thanks :)
#557 is merged now.

fmuellner:
Noticed :-)

Pushed an updated branch, the newly added tests should work no (at least they did in a local test)

bmeneguele:
@fmuellner I just noticed that the gitlab API doesn't offer a way to "follow" the pipeline jobs URL in case it lives in another repo other than source or target, right? :-(.

bmeneguele (approved):
Considering the API limitation I mentioned above the MR looks good to me.

fmuellner:
> I just noticed that the gitlab API doesn't offer a way to "follow" the pipeline jobs URL in case it lives in another repo other than source or target, right? :-(.

Do you mean https://docs.gitlab.com/ee/ci/multi_project_pipelines.html? I don't have any experience with that, or know anything about other ways how jobs can end up in another repo.

But considering that merge-request pipelines (which can run on either source- or target repo  - my guess is that it's based on permissions, i.e. whether the MR owner has developer permissions on the target) doesn't expose that information in the API, I'm not optimistic ...

bmeneguele:
> 
> 
> > I just noticed that the gitlab API doesn't offer a way to "follow" the pipeline jobs URL in case it lives in another repo other than source or target, right? :-(.
> 
> Do you mean https://docs.gitlab.com/ee/ci/multi_project_pipelines.html? I don't have any experience with that, or know anything about other ways how jobs can end up in another repo.

Ah, ok! I wasn't aware of that. I'm going to take a look!

> 
> But considering that merge-request pipelines (which can run on either source- or target repo - my guess is that it's based on permissions, i.e. whether the MR owner has developer permissions on the target) doesn't expose that information in the API, I'm not optimistic ...

Hmm, good point. I'm going to take a closer look on that too.
But anything that comes out of this is just enhancement to the feature this MR enables.
So.. I'm going to merge it :)

PR #539, opened 2020-12-22
mr_list: fix latest updated MR test

bmeneguele:
Update the latest MR name to which mr_list_test checks.


prarit (approved):

bmeneguele:
Ok, it's a bit trickier than I thought.
We have some tests searching for latest updated MR, which !19 is still the correct, since that's what we have running for every PR, but not !22 since it's not upstreamed yet.
However, we also have those tests looking for the latest created MR, which will be !22.

I'm going to create another var, so we can have both tracked in different ways.

Issue #406, opened 2020-08-27
Migrate to GitLab


zaquestion:
GitLab has had an open issue for a while now desiring a "Lab CLI for GitLab", and we've made ourselves known. More recently we've started talking about officially making the move into GitLab's official namespace. 

See: https://gitlab.com/gitlab-org/gitlab/-/issues/14448#note_399479784 

This issue aims to capture the immediate todos to get `lab` fully re-homed functionally speaking. (no particular order)

- [ ] Replace Travis CI with GitLab CI
  - [ ] Update `.goreleaser` to upload artifacts as GitLab release artifacts
  - [x] Update homebrew tap home? Submit formula to homebrew core?
    - [x] https://github.com/zaquestion/homebrew-tap (probably going to want to update this going forward for a while, but should also setup a new one, or perhaps automation to `homebrew-core`
- [ ] Update `install.sh` to look for GitLab homed artifacts
- [ ] Host https://zaquestion.github.io/lab/ (`./docs`) in GitLab Pages
- [ ] Update tests (some rely on existing `GOPATH`)
  - [ ] Deal with https://gitlab.com/zaquestion/test
- [ ] Add existing collaborators to new repo
  - [ ] https://gitlab.com/zaquestion
  - [ ] https://gitlab.com/prarit
  - [ ] https://gitlab.com/rsteube
  - [ ] https://gitlab.com/claytonrcarter (if he's still interested?)
- [ ] Get `0.18.0` out (good forcing function for this stuff)

Longer Term
- [ ] @zaquestion: go through all open issues and organize, label, rank, etc
- [ ] Start a `v1` branch so gitlabers and `lab` maintainers alike can start introducing new vision, rehauls, and other breaking changes (Starting with items in the [1.0 project](https://github.com/zaquestion/lab/projects/2))

zaquestion:
@rsteube @prarit @claytonrcarter for visibility

(y'alls opinions are welcome too, if there's any opposition in moving over nows the time to voice it ;)

prarit:
I have a copy of lab on gitlab FWIW.  https://gitlab.com/prarit/lab 

I will nuke it if necessary to free up the namespace.

rsteube:
Nah, makes sense as it will tend to more usage of it during development (EYODF).
Keeping track will be a bit harder for me though as everything else is on github so far.
Might want to get #151 done first so that it won't interfere with any work profiles.
[...]

PR #540, opened 2020-12-23
ci_view: Fix screen freeze

### LINKED ISSUE #538: lab ci view does not react to key input
I have built the latest git version (fd0096293383eb736ab7431dc0a1e2d2f2144f29); when I run `lab ci view`, the pipeline shows up in the overview.

Key input, however, does not show any action. I have to use a different shell to `killall lab` to quit. I use it with alacritty, TERM=xterm-color256, but also tested xterm & TERM=xterm - no change.

fmuellner:
After the tcell/v2 update, any key press during `lab ci view` results
in a lock up. It is unclear whether the changes in input handling on
the tcell side are intentional, but we can address the issue by making
the channel that notifies about input events asynchronous and ensure
that we don't overrun its queue.

Fixes #538

### DIFF
--- a/cmd/ci_view.go
+++ b/cmd/ci_view.go
@@ -72,7 +72,7 @@ Feedback Encouraged!: https://github.com/zaquestion/lab/issues`,
 
 		boxes = make(map[string]*tview.TextView)
 		jobsCh := make(chan []*gitlab.Job)
-		inputCh := make(chan struct{})
+		inputCh := make(chan struct{}, 1)
 
 		screen, err := tcell.NewScreen()
 		if err != nil {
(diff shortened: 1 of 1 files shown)


bmeneguele (approved):

zaquestion:
@davinkevin does this fix the freeze / crash you mentioned in #406

davinkevin:
Yeah, this is working well now 👍. So, you can forget my comment in #406. 

BTW, I will use the compiled binary I have because it has a lot of fixes and this is way simpler to use now (especially on the CI view…).

Thank @zaquestion && @fmuellner 😉

PR #541, opened 2020-12-23
util: Support getting MR ID from branch name

fmuellner:
If no MR ID is specified on the command line, the mr subcommands fall
back to looking for a MR associated with the current branch. Extend
that behavior to also try a branch name specified on the command line,
as branch names are usually easier to remember than MR numbers.

### DIFF
--- a/cmd/mr_edit.go
+++ b/cmd/mr_edit.go
@@ -22,36 +22,27 @@ var mrEditCmd = &cobra.Command{
 	Long:    ``,
 	Example: `lab MR edit <id>                                # update MR via $EDITOR
 lab MR update <id>                              # same as above
+lab MR update <branch-name>                     # same, but get MR ID from local branch
 lab MR edit <id> -m "new title"                 # update title
 lab MR edit <id> -m "new title" -m "new desc"   # update title & description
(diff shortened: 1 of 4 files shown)


prarit:
Ah this was on my list of TODOs :)

cmd/mr_edit.go
```diff
@@ -22,36 +22,27 @@ var mrEditCmd = &cobra.Command{
 	Long:    ``,
 	Example: `lab MR edit <id>                                # update MR via $EDITOR
 lab MR update <id>                              # same as above
+lab MR update my-branch                         # same, but get MR ID from local branch
```
prarit, line 25: Nitpick -- maybe "my-branch-name" ?  Only change this if I find something else to complain about :)

prarit (approved):
LGTM

cmd/mr_edit.go
bmeneguele: I would even suggest `<branch-name>`, but I know there are other help messages around that don't make use of `<>`, so... :))

bmeneguele (approved):

cmd/mr_edit.go
fmuellner: Right, that would be more consistent with the `<id>` bit right above.

PR #542, opened 2020-12-23
Improve comments appearance

fmuellner:
A couple of minor fixes and tweaks, and a bigger one: Render markdown in comments as we do for the desciption.

The last change means that only the header line of new comments is printed in bold, but IMHO this is enough for them to stand out.

And the readability improvement is quite significant:

Screenshot from 2020-12-23 12-51-08

### DIFF
--- a/cmd/issue_note_test.go
+++ b/cmd/issue_note_test.go
@@ -80,7 +80,8 @@ func Test_issueReplyNote(t *testing.T) {
 	noteID := noteIDs[1]
 
 	// add reply to the noteID
-	reply := exec.Command(labBinaryPath, "issue", "reply", "lab-testing", issueID+":"+noteID, "-m", "reply to note")
+	reply := exec.Command(labBinaryPath, "issue", "reply", "lab-testing", issueID+":"+noteID,
+		"-m", "reply to note", "-m", "second reply paragraph")
 	reply.Dir = repo
 	c, err := reply.CombinedOutput()
 	if err != nil {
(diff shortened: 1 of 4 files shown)


prarit:
I had originally implemented only bolding the new comment string like you did but found that it was useful to have the whole message bolded with MRs with lots of comments.

I'm on the fence on this.  I'll leave it @bmeneguele or @zaquestion to make a decision on this one.

prarit:
One thought: It might be nice to have an MR format test of some kind.  Some of your commits are for indentation and readability, and it might be good to have a test that checks for that the output format is correct.

bmeneguele:
> I had originally implemented only bolding the new comment string like you did but found that it was useful to have the whole message bolded with MRs with lots of comments.
> 
> I'm on the fence on this. I'll leave it @bmeneguele or @zaquestion to make a decision on this one.

IMHO it's better to keep to whole comment bold, mostly due to new comments in old discussions that are shown before newer comments in that MR. But the other changes you did indeed look pretty nice. Is it possible to keep both? :)

fmuellner:
>  But the other changes you did indeed look pretty nice. Is it possible to keep both? :)

My first reaction was to say no, because `printit()` is still set up to use bold style, but the rendered markdown overrides it.

But after taking a closer look, I think it'll be possible to come up with something.

fmuellner:
> One thought: It might be nice to have an MR format test of some kind. Some of your commits are for indentation and readability, and it might be good to have a test that checks for that the output format is correct.

Good point, I'll amend the tests to check for the indent.

bmeneguele (approved):
Yep! LGTM :)

PR #546, opened 2020-12-28
RFC: Figure out issue automatically if possible

fmuellner:
If no issue number is specified on the command line, we may still
get one if we find a MR request associated with the current branch,
and that MR is set to close an issue.

This is a follow-up from #541, but I kept it out of that pull request
because I'm much less convinced that it's a good idea. It *can* be convenient,
but also surprising and confusing ...

### DIFF
--- a/cmd/issue_browse.go
+++ b/cmd/issue_browse.go
@@ -20,7 +20,7 @@ var issueBrowseCmd = &cobra.Command{
 	Long:             ``,
 	PersistentPreRun: LabPersistentPreRun,
 	Run: func(cmd *cobra.Command, args []string) {
-		rn, num, err := parseArgsRemoteAndID(args)
+		rn, num, err := parseArgsRemoteAndIssueID(args)
 		if err != nil {
 			log.Fatal(err)
 		}
--- a/cmd/issue_close.go
+++ b/cmd/issue_close.go
@@ -22,7 +22,7 @@ var issueCloseCmd = &cobra.Command{
 lab issue close --duplicate 123 1234
 lab issue close --duplicate other-project#123 1234`,
 	Run: func(cmd *cobra.Command, args []string) {
-		rn, id, err := parseArgsRemoteAndID(args)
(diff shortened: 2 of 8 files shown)


bmeneguele:
If an MR has a relation to two different issues the first one returned by the API will be used, right?
Yes, it can be quite confusing indeed. And also, showing the content for both issues would be even confusing.

What if, instead of actually presenting any content, it first looks for the number of related issues, in case it's >1 then a similar (or exactly the same) output of `lab issue list` is presented with additional words like: `please choose one from the list`. Otherwise (number of issues == 1), present the content of that specific issue.

If that's too much trouble, I would vote for letting this feature out.

fmuellner:
> What if, instead of actually presenting any content, it first looks for the number of related issues, in case it's >1 then a similar (or exactly the same) output of `lab issue list` is presented with additional words like: `please choose one from the list`, then the user can pick one and pass it explicitly to `lab issue show <id>`. Otherwise (number of related issues == 1), present the content of that specific issue.

I like that lab is generally non-interactive (in contrast for instance to github's new cli tool), but doing it only in the >1 case sounds good to me.


> I would vote for letting this feature out.

I'd also be fine with that to be honest.

bmeneguele:
Your call then ;).
Feel free to close this PR in case you 

> > What if, instead of actually presenting any content, it first looks for the number of related issues, in case it's >1 then a similar (or exactly the same) output of `lab issue list` is presented with additional words like: `please choose one from the list`, then the user can pick one and pass it explicitly to `lab issue show <id>`. Otherwise (number of related issues == 1), present the content of that specific issue.
> 
> I like that lab is generally non-interactive (in contrast for instance to github's new cli tool), but doing it only in the >1 case sounds good to me.
> 

Yes, I feel the same about `lab` non-interactiveness.

> > I would vote for letting this feature out.
> 
> I'd also be fine with that to be honest.

Lets keep it out then :), avoiding possibly dead code for future features.
Unless, of course, @zaquestion and/or @prarit have another opinion.

PR #547, opened 2020-12-28
cmd: Pipe output through pager in terminals

fmuellner:
Replicate git's behavior of piping output through a pager when it
doesn't fit a single screen. This is not only more convenient than
piping into less on the command line, but preserves terminal features
like colored output as well.

I do have a local follow-up to add a global --no-pager option like git
as well, if that's wanted (not sure that's the case, so leaving it out
right now)

### DIFF
--- a/cmd/ci_status.go
+++ b/cmd/ci_status.go
@@ -43,6 +43,9 @@ lab ci status --wait`,
 
 		pid := rn
 
+		pager := NewPager(cmd.Flags())
+		defer pager.Close()
+
 		w := tabwriter.NewWriter(os.Stdout, 2, 4, 1, byte(' '), 0)
 
 		wait, err := cmd.Flags().GetBool("wait")
--- a/cmd/ci_trace.go
+++ b/cmd/ci_trace.go
@@ -53,6 +53,9 @@ var ciTraceCmd = &cobra.Command{
 		}
 		projectID = project.ID
 
+		pager := NewPager(cmd.Flags())
+		defer pager.Close()
+
 		err = doTrace(context.Background(), os.Stdout, projectID, pipelineID, jobName)
 		if err != nil {
 			log.Fatal(err)
--- a/cmd/issue_list.go
+++ b/cmd/issue_list.go
@@ -37,6 +37,10 @@ lab issue list remote "search terms"  # search "remote" for issues with "search 
 		if err != nil {
 			log.Fatal(err)
 		}
+
+		pager := NewPager(cmd.Flags())
+		defer pager.Close()
+
 		for _, issue := range issues {
 			fmt.Printf("#%d %s\n", issue.IID, issue.Title)
 		}
--- a/cmd/issue_show.go
+++ b/cmd/issue_show.go
@@ -42,6 +42,9 @@ var issueShowCmd = &cobra.Command{
 			renderMarkdown = !noMarkdown
 		}
 
+		pager := NewPager(cmd.Flags())
+		defer pager.Close()
+
 		printIssue(issue, rn, renderMarkdown)
 
 		showComments, _ := cmd.Flags().GetBool("comments")
--- a/cmd/label_list.go
+++ b/cmd/label_list.go
@@ -35,6 +35,9 @@ lab label list remote "search term"  # search "remote" for labels with "search t
 			log.Fatal(err)
 		}
 
+		pager := NewPager(cmd.Flags())
+		defer pager.Close()
+
 		for _, label := range labels {
 			// GitLab API has no search for labels, so we do it ourselves
 			if labelSearch != "" &&
--- a/cmd/mr_list.go
+++ b/cmd/mr_list.go
@@ -50,6 +50,10 @@ lab mr list remote "search terms"  # search "remote" for merge requests with "se
 			log.Print(err)
 			config.UserConfigError()
 		}
+
+		pager := NewPager(cmd.Flags())
+		defer pager.Close()
+
 		for _, mr := range mrs {
 			fmt.Printf("!%d %s\n", mr.IID, mr.Title)
 		}
--- a/cmd/mr_show.go
+++ b/cmd/mr_show.go
@@ -65,6 +65,9 @@ var mrShowCmd = &cobra.Command{
 			}
 			git.Show(remote+"/"+mr.TargetBranch, mr.SHA, mrShowPatchReverse)
 		} else {
+			pager := NewPager(cmd.Flags())
+			defer pager.Close()
+
 			printMR(mr, rn, renderMarkdown)
 		}
 
--- a/cmd/project_list.go
+++ b/cmd/project_list.go
@@ -48,6 +48,10 @@ var projectListCmd = &cobra.Command{
 		if err != nil {
 			log.Fatal(err)
 		}
+
+		pager := NewPager(cmd.Flags())
+		defer pager.Close()
+
 		for _, p := range projects {
 			fmt.Println(p.PathWithNamespace)
 		}
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -102,6 +102,7 @@ func init() {
 	RootCmd.SetHelpCommand(helpCmd)
 	RootCmd.SetHelpFunc(helpFunc)
 	RootCmd.Flags().Bool("version", false, "Show the lab version")
+	RootCmd.PersistentFlags().Bool("no-pager", false, "Do not pipe output into a pager")
 }
 
 var (
--- a/cmd/snippet_list.go
+++ b/cmd/snippet_list.go
@@ -28,6 +28,8 @@ var snippetListCmd = &cobra.Command{
 		if err != nil {
 			log.Fatal(err)
 		}
+		pager := NewPager(cmd.Flags())
+		defer pager.Close()
 		for _, snip := range snips {
 			fmt.Printf("#%d %s\n", snip.ID, snip.Title)
 		}
--- a/cmd/util.go
+++ b/cmd/util.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"log"
 	"os"
+	"os/exec"
 	"strconv"
 	"strings"
 	"syscall"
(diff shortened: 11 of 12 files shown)


bmeneguele:
It seems the test code should be slightly changed :-)
But I like the idea and the code is also looking good IMO.
Now, about the `--no-pager`: even though I would be fine without it and, tbh, I'm not sure when people would really use it, I think it's easy enough to implement. So... if you have a real use case, go for it. Otherwise, let's keep it out for now.

bmeneguele:
BTW, I fixed the error with do_fork_name_path (or at least it should be fixed).

fmuellner:
> It seems the test code should be slightly changed :-)

Oh, indeed - that was a bit buried in the fork test failures you fixed :-)

Should be fixed with #550 now.

> Now, about the `--no-pager`: even though I would be fine without it and, tbh, I'm not sure when people would really use it, I think it's easy enough to implement. So... if you have a real use case, go for it. Otherwise, let's keep it out for now.

I don't have a use case. Presumably there is one for some people because git has the option, but I don't remember using it ever.

bmeneguele:
> > It seems the test code should be slightly changed :-)
> 
> Oh, indeed - that was a bit buried in the fork test failures you fixed :-)
> 

Yeah, it was a left over project in our lab-testing repository created by a test I created when working on the fork+project creation name/namespace/path support. But I didn't dig enough to find the root cause where that repo wasn't deleted when it should have.
I'll keep my eyes on that.

> Should be fixed with #550 now.
> 

Ack, thanks!

> > Now, about the `--no-pager`: even though I would be fine without it and, tbh, I'm not sure when people would really use it, I think it's easy enough to implement. So... if you have a real use case, go for it. Otherwise, let's keep it out for now.
> 
> I don't have a use case. Presumably there is one for some people because git has the option, but I don't remember using it ever.

You said you already have the code ready, right? Go for it then.
Otherwise, I would say to "not worry about it" :-).

fmuellner:
> You said you already have the code ready, right? Go for it then.
> Otherwise, I would say to "not worry about it" :-).

Mmh, OK.

To be honest, I wasn't sure how the pager support would be received, and so did the patch mainly to have the "but you can disable it" argument at hand :-)

bmeneguele:
> > You said you already have the code ready, right? Go for it then.
> > Otherwise, I would say to "not worry about it" :-).
> 
> Mmh, OK.
> 
> To be honest, I wasn't sure how the pager support would be received, and so did the patch mainly to have the "but you can disable it" argument at hand :-)

Got it :).
From my pov, using a pager is really useful + helpful, mostly when you have big discussions going on and a terminal that doesn't help you a lot (or you prefer to use the common keyboard keys to navigate instead of pressing mouse buttons or PgDn + PgUp => that's my case using xterm). Soo.. I really appreciated it.

bmeneguele (approved):

prarit:
I like this and just tested with a pipe and redirect.  It looks like things work except that I have an odd bug.  I can't seem to quit the pager.  When I page with less normally, I can enter 'q' to quit less.  With this use of less, however, 'q' does not work.  I can see that I'm using /usr/bin/less for the pager in 'ps aux':

```
prarit    404509  0.0  0.0 221024  2536 pts/2    S+   12:43   0:00 /usr/bin/less
prarit    404527  5.1  0.1 2252944 81588 pts/6   Sl+  12:43   0:00 /home/prarit/git-kernel/github/lab/lab mr show 51 --comments
prarit    404555  0.0  0.0 221024  2424 pts/6    S+   12:43   0:00 less
```

@fmuellner is 'q' to quit less working for you?

fmuellner:
> I can't seem to quit the pager unless I use CTRL-C. When I page with less normally, I can enter 'q' to quit less. With this use of less, however, 'q' does not work

Weird. I can't think of a reason for that, it's certainly not something I have seen with the patch applied.



> I can see that I'm using /usr/bin/less for the pager in 'ps aux':
> 
> ```
> prarit    404509  0.0  0.0 221024  2536 pts/2    S+   12:43   0:00 /usr/bin/less
> prarit    404527  5.1  0.1 2252944 81588 pts/6   Sl+  12:43   0:00 /home/prarit/git-kernel/github/lab/lab mr show 51 --comments
> prarit    404555  0.0  0.0 221024  2424 pts/6    S+   12:43   0:00 less
> ```

This does look a bit strange: I would expect the second `less` to be the pager here, as it's created after the `lab` process and shares the same TTY.

fmuellner:
> @fmuellner is 'q' to quit less working for you?

Wow, turns out I was too quick to answer. It's bizarre, it works after having scrolled to the bottom, but is broken before that (after scrolling down - for example with `G`, the position doesn't matter anymore; `q` still work when I jump back to the top with `g` for instance).

fmuellner:
Two more observations:
 - `q` actually works fine as far as less is concerned; the problem is that for some reason `PrintDiscussions()` gets stuck, thus the pager's `Close()` method isn't called and we don't `Wait()` on the pager process and less ends up as a zombie
 - this doesn't happen when using the `--no-markdown` flag

bmeneguele:
That's actually strange. I thought I had overlooked the issue, but it actually doesn't happen with me.
The pager returned by `git.PagerCommand()` here was `less` and I'm using `alacritty` as my terminal.
Can it be somehow related to the console program being used (xterm, xfce-term, konsole, gnome-terminal, ...)?

fmuellner:
> Can it be somehow related to the console program being used (alacrity, xterm, xfce-term, konsole, gnome-terminal, ...)?

Oh, I hadn't considered that. It's at least a possibility.

bmeneguele:
xterm didn't fail for me as well. what terminals are you both @fmuellner and @prarit using?

fmuellner:
Mmh, interesting. I'm usually using `gnome-terminal`, but did a quick test in `xterm` which did reproduce the issue as well.

Maybe not terminal-dependent after all, and something in the environment instead?

bmeneguele:
> Maybe not terminal-dependent after all, and something in the environment instead?

Yeah, probably :/.
You said that `PrintDiscussion` was the point where lab got stuck right?
I'll try to dig a bit there and hopefully find a way to reproduce the issue here.
I have a second machine lurking around here at home, will have a try there too.

bmeneguele:
Any updates in here?
I wasn't able to reproduce the issue in any of my systems =(.

fmuellner:
I'm afraid not.

(I'm out of ideas of what to look for, so didn't spend any time on this)

fmuellner:
Small update: Earlier today I was surprised by the pager *not* getting stuck. I did some experimentation on a test issue, and the issue appears to be related to the length of the discussion (I tried different markup patterns without breaking, then kept adding "lorem ipsum" comments until it did).

No closer to a fix though :-(

prarit:
I wonder if this has something to do with the way we're outputting the messages.  Is the for loop to blame -- ie) is the code stuck in that for loop and not answering to external output?

prarit:
Here's something to try : Dump all the text to a buffer first, and then dump it to screen with the pager functionality.  That would prove/disprove my for loop theory.

fmuellner:
Doesn't appear to make a difference.

In case it helps, this is the backtrace I get in gdb when in the stuck state (without the changes to print to a string first):
```
Thread 8 (Thread 0x7fffc9bb0640 (LWP 749589)):
#0  syscall.Syscall () at /usr/lib/golang/src/syscall/asm_linux_amd64.s:24
#1  0x00000000004bea5a in syscall.write (fd=8, p=[]uint8, n=<optimized out>, err=...) at /usr/lib/golang/src/syscall/zsyscall_linux_amd64.go:914
#2  0x00000000004bcc09 in syscall.Write (fd=8, p=[]uint8 = {...}, n=<optimized out>, err=...) at /usr/lib/golang/src/syscall/syscall_unix.go:212
#3  0x00000000004dd012 in internal/poll.ignoringEINTR (fn={void (int, []uint8, int *, error *)} 0xc000946b00, fd=8, p=[]uint8 = {...}, ~r3=<optimized out>, ~r4=...) at /usr/lib/golang/src/internal/poll/fd_unix.go:567
#4  0x00000000004dafac in internal/poll.(*FD).Write (fd=0xc0001b62a0, p=[]uint8 = {...}, ~r1=<optimized out>, ~r2=...) at /usr/lib/golang/src/internal/poll/fd_unix.go:267
#5  0x00000000004e2317 in os.(*File).write (f=0xc0001b4120, b=[]uint8 = {...}, n=<optimized out>, err=...) at /usr/lib/golang/src/os/file.go:173
#6  os.(*File).Write (f=0xc0001b4120, b=[]uint8 = {...}, n=<optimized out>, err=...) at /usr/lib/golang/src/os/file.go:173
#7  0x00000000004ed8c5 in fmt.Fprintf (w=..., format="%s\n", a=[]interface {} = {...}, n=<optimized out>, err=...) at /usr/lib/golang/src/fmt/print.go:205
#8  0x0000000000dbd98b in github.com/fatih/color.(*Color).Printf (c=0xc000913640, format="%s\n", a=[]interface {} = {...}, n=0, err=...) at /var/home/fmuellner/go/pkg/mod/github.com/fatih/color@v1.10.0/color.go:228
#9  0x0000000000dbeedc in github.com/fatih/color.(*Color).PrintfFunc.func1 (format="%s\n", a=[]interface {} = {...}) at /var/home/fmuellner/go/pkg/mod/github.com/fatih/color@v1.10.0/color.go:297
#10 0x0000000000ea7e51 in github.com/zaquestion/lab/cmd.PrintDiscussions (discussions=[]*github.com/xanzy/go-gitlab.Discussion, since="", idstr="mr", idNum=1572, renderMarkdown=true) at /var/home/fmuellner/src/lab/cmd/issue_show.go:350
#11 0x0000000000ec4345 in github.com/zaquestion/lab/cmd.glob..func40 (cmd=0x18d2000, args=[]string) at /var/home/fmuellner/src/lab/cmd/mr_show.go:86
#12 0x00000000005d55c2 in github.com/spf13/cobra.(*Command).execute (c=0x18d2000, a=[]string = {...}, err=...) at /var/home/fmuellner/go/pkg/mod/github.com/spf13/cobra@v1.1.1/command.go:854
#13 0x00000000005d62f5 in github.com/spf13/cobra.(*Command).ExecuteC (c=0x18d2000, cmd=<optimized out>, err=...) at /var/home/fmuellner/go/pkg/mod/github.com/spf13/cobra@v1.1.1/command.go:958
#14 0x0000000000eb3da8 in github.com/spf13/cobra.(*Command).Execute (c=0xfffffffffffffe00) at /var/home/fmuellner/go/pkg/mod/github.com/spf13/cobra@v1.1.1/command.go:895
#15 github.com/zaquestion/lab/cmd.Execute () at /var/home/fmuellner/src/lab/cmd/root.go:222
#16 0x0000000000eca553 in main.main () at /var/home/fmuellner/src/lab/main.go:28
```

fmuellner:
Looks like switching to the low-level StartProcess API works.

zaquestion:
Looking solid folks, and for what it's worth, I appreciate the addition of a `--no-pager` option :) like with `git`, it's useful in cases where `lab` is being used with automation. Always appreciate changes to make `lab` feel more like `git`

bmeneguele:
@prarit could you re-test this feature with the new push @fmuellner did using a lower-level process API?

prarit (approved):
Looks good to me.  Thanks for fixing this!

bmeneguele (approved):
Well done @fmuellner !

PR #548, opened 2020-12-28
labels: Support partial label names

fmuellner:
Currently label names have to be specified by their full name, and
any partial or non-existing names end up creating new labels. Instead,
map the list of names we get on the command line to the real labels.

This allows user-friendly (read: long and descriptive) label names
without losing command line convenience.

### DIFF
--- a/cmd/issue_create.go
+++ b/cmd/issue_create.go
@@ -34,7 +34,7 @@ var issueCreateCmd = &cobra.Command{
 		if err != nil {
 			log.Fatal(err)
 		}
-		labels, err := cmd.Flags().GetStringSlice("label")
+		labelTerms, err := cmd.Flags().GetStringSlice("label")
 		if err != nil {
 			log.Fatal(err)
 		}
@@ -53,6 +53,11 @@ var issueCreateCmd = &cobra.Command{
 			}
 		}
 		rn, err := git.PathWithNameSpace(remote)
+		if err != nil {
+			log.Fatal(err)
+		}
+
+		labels, err := MapLabels(rn, labelTerms)
 		if err != nil {
 			log.Fatal(err)
 		}
--- a/cmd/issue_edit.go
+++ b/cmd/issue_edit.go
@@ -67,13 +67,21 @@ lab issue edit <id>:<comment_id>                   # update a comment on MR`,
 		}
 
 		// get the labels to add
-		labels, err := cmd.Flags().GetStringSlice("label")
+		labelTerms, err := cmd.Flags().GetStringSlice("label")
+		if err != nil {
+			log.Fatal(err)
+		}
+		labels, err := MapLabels(rn, labelTerms)
 		if err != nil {
 			log.Fatal(err)
 		}
 
 		// get the labels to remove
-		unlabels, err := cmd.Flags().GetStringSlice("unlabel")
+		unlabelTerms, err := cmd.Flags().GetStringSlice("unlabel")
+		if err != nil {
+			log.Fatal(err)
+		}
+		unlabels, err := MapLabels(rn, unlabelTerms)
 		if err != nil {
 			log.Fatal(err)
 		}
--- a/cmd/issue_edit_test.go
+++ b/cmd/issue_edit_test.go
@@ -70,7 +70,7 @@ func Test_issueEditLabels(t *testing.T) {
 
 	// update the issue
 	cmd := exec.Command(labBinaryPath, "issue", "edit", "lab-testing", issueNum,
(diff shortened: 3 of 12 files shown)


bmeneguele:
Doesn't it mess with workflows where a label is indeed a substring of another label?
For instance:
1) a user really wants to create a label named `DOC` meaning something different from `documentation`: in this case `lab` will set `documentation` to the issue instead of creating `DOC`.
2) a user creates `DOC` on webui and the other user using `lab` suddenly starts to see his workflow break because he was using `doc` for `documentation`.

I might be overthinking here, but is this support really required? Labels are meant to be short and objective unless the project is using "scoped labels", but in this case, a full label name must be used anyway.

fmuellner:
> Doesn't it mess with workflows where a label is indeed a substring of another label?

Mmh, yes. Maybe it works if we prioritize exact matches (and don't complain about the ambiguity in that case)?

> 1. a user really wants to create a label named `DOC` meaning something different from `documentation`: in this case `lab` will set `documentation` to the issue instead of creating `DOC`.

For creating a new label, an explicit `label create` command would be more intuitive IMHO than just creating labels on-the-fly. And the name overlap could be addressed by matching on the exact name as well (or maye figuring out that `DOC` is a better match for `doc` than `documentation`)

> 2\. a user creates `DOC` on webui and the other user using `lab` suddenly starts to see his workflow break because he was using `doc` for `documentation`.

Not sure how realistic that case is, but that's a downside that I don't think can be addressed.

> is this support really required?

Not *required*, but those labels are quite a PITA:
https://gitlab.gnome.org/groups/GNOME/-/labels

At the very least I'd like to avoid accidental label creation when typing `2. Needs information` instead of `2. Needs Information`.

bmeneguele:
> > Doesn't it mess with workflows where a label is indeed a substring of another label?
> 
> Mmh, yes. Maybe it works if we prioritize exact matches (and don't complain about the ambiguity in that case)?
> 

Yes! +1 for this idea.

> > 1. a user really wants to create a label named `DOC` meaning something different from `documentation`: in this case `lab` will set `documentation` to the issue instead of creating `DOC`.
> 
> For creating a new label, an explicit `label create` command would be more intuitive IMHO than just creating labels on-the-fly. And the name overlap could be addressed by matching on the exact name as well (or maye figuring out that `DOC` is a better match for `doc` than `documentation`)
> 

Good point and agreed 100%.

> > 2. a user creates `DOC` on webui and the other user using `lab` suddenly starts to see his workflow break because he was using `doc` for `documentation`.
> 
> Not sure how realistic that case is, but that's a downside that I don't think can be addressed.
> 

Yea, you're right, I overthought it.
A user with enough access to set labels are usually the ones managing them.

> > is this support really required?
> 
> Not _required_, but those labels are quite a PITA:
> https://gitlab.gnome.org/groups/GNOME/-/labels
> 
> At the very least I'd like to avoid accidental label creation when typing `2. Needs information` instead of `2. Needs Information`.

Ahh.. ok.. I feel the pain now :-).
I think that for this PR just adding the "exact match first" approach is enough.

Then, a follow-up for `lab label create` would be really welcome IMHO as well. 
Would you like to take it too? (if you're swamped I can take over it, no worries at all).

fmuellner:
> Then, a follow-up for `lab label create` would be really welcome IMHO as well.
> Would you like to take it too? (if you're swamped I can take over it, no worries at all).

I'm on PTO until Monday, that should be enough time to implement a `create` subcommand :-)

rsteube:
Adding completion for labels is rather simple if that helps: https://github.com/zaquestion/lab/compare/master...rsteube:add-label-completion

bmeneguele:
> Adding completion for labels is rather simple if that helps: [master...rsteube:add-label-completion](https://github.com/zaquestion/lab/compare/master...rsteube:add-label-completion)

label completion is something indeed interesting to have a look, maybe worth opening a new PR after this one is merged?

bmeneguele (approved):
LGTM

bmeneguele:
The test failed doesn't really have relation to the code here being touched.
And I think it was already solved in the lab-testing tree.
@fmuellner I'm not being able to re-run the TravisCI build, could you try?

fmuellner:
> I'm not being able to re-run the TravisCI build, could you try?

Mmh, I pushed a rebased branch for that, and somehow the tests are gone now?!

bmeneguele:
> Mmh, I pushed a rebased branch for that, and somehow the tests are gone now?!

LOL.. did we broke github? Weird..
I'm going to merge it anyway. In case the build fails in the next PR we work on that then.

PR #552, opened 2021-01-04
place code used in multiple places in common files

bmeneguele:
Today, some functions and variables are used in different places but are declared in a specific command file and shared elsewhere. It's not only confusing for the developer trying to understand the code but also is error-prone: code duplication is really easy if the developer isn't aware that other commands may already use something similar.

The idea of this patchset is to place "common" code in "common" files based on their purpose. For instance: common code used in both `mr_edit.go` and `issue_edit` that originally was declared in `issue_edit.go` is now placed in a `edit_common.go` file. Also, this patchset doesn't bring any code behavior changes. Some tests were moved from one file to another, but all should pass as usual, since nothing was changed.

This is still a RFC, so please, "help wanted" :). I'm not sure how acceptable or good is this approach, maybe a different file naming would be better? But I'm sure this code organization is a must, even more with the features growing quickly. Even though I'm a really bad programmer, I always get annoyed when I'm forced to declare a variable in one file while other variable that shares the "same" context is declared in another command file (the same happens for functions). The Go packaging encapsulation scheme doesn't really help (it's like using C `extern` all over the places), that's why I think we should start re-organizing the code. 

Also, I know that this code will break some of the opened PRs, so I don't really plan to merge it too soon, I'm going to, at least, wait for the PRs that are open now to get merged/closed. And, hopefully, I've found every "common" code existent, otherwise, we can move the leftovers as we touch them.


zaquestion:
I'm favorable towards these refactors. 

Rough Thoughts: 

Waaayyy back in the day we discussed breaking the commands `mr`, `issue`, etc into their on packages within `cmd` (i.e. `cmd/mr`), I can imagine that being painful/surgical to even achieve.... 

For the first step I think moving code around within the `cmd` package is the best approach. Longer term shared functionality might make more sense in as a package in `internal`.

bmeneguele:
> I'm favorable towards these refactors.
> 
> Rough Thoughts:
> 
> Waaayyy back in the day we discussed breaking the commands `mr`, `issue`, etc into their on packages within `cmd` (i.e. `cmd/mr`), I can imagine that being painful/surgical to even achieve....

Yes, I think that's indeed a good path to follow: each sub command having its own and self contained package.
I'm going to open an issue about that so we can further discuss about the project structure goal.

> For the first step I think moving code around within the `cmd` package is the best approach. Longer term shared functionality might make more sense in as a package in `internal`.

yes, probably. even more if we have the commands being splitted in different folders.

prarit:
Me too.  When I merged the note code I was thinking there must be a better way to do this.  Using _common.go files is definitely a move the right direction.

fmuellner:
> Also, I know that this code will break some of the opened PRs, so I don't really plan to merge it too soon

Please don't hold back for any of my stuff, I'm happy to rebase as necessary.

bmeneguele:
> > Also, I know that this code will break some of the opened PRs, so I don't really plan to merge it too soon
> 
> Please don't hold back for any of my stuff, I'm happy to rebase as necessary.

I think it's better to indeed get those merged first, avoiding revamping the patches that were already written for some time :).

bmeneguele:
The code is essentially the same, I just rebased it and added another commit for moving some code added by one of the  @fmuellner PRs related to CI.

PR #557, opened 2021-01-06
mr_list: Update test (again)

fmuellner:
!22 got merged, so no longer shows up in the list by default.


bmeneguele (approved):
Thanks Florian!

Issue #558, opened 2021-01-07
lab 0.18.0 behavior change


chenrui333:
👋  looks like the lab 0.18.0 behavior got changed (it requires the gitlab PAT before proceeding to do something meaningful) and now the test started not working.

```
    ENV["LAB_CORE_USER"] = "test_user"
    ENV["LAB_CORE_HOST"] = "https://gitlab.com"
    ENV["LAB_CORE_TOKEN"] = "dummy"
    ENV["GIT_AUTHOR_NAME"] = "test user"
    ENV["GIT_AUTHOR_EMAIL"] = "test@example.com"
    ENV["GIT_COMMITTER_NAME"] = "test user"
    ENV["GIT_COMMITTER_EMAIL"] = "test@example.com"
    system "git", "init"
    %w[haunted house].each { |f| touch testpath/f }
    system "git", "add", "haunted", "house"
    system "git", "commit", "-a", "-m", "Initial Commit"
    assert_equal "haunted\nhouse", shell_output("#{bin}/lab ls-files").strip
```

Interesting enough, I found the similar [test code](https://github.com/zaquestion/lab/blob/0ff2ff02b4ecffbbeb1f695556486d104d30f56e/.goreleaser.yml#L20-L25) in the codebase. Not sure if some real test token was used in the goreleaser process.

relates to https://github.com/Homebrew/homebrew-core/pull/68377

bmeneguele:
Hi @chenrui333 , thanks for opening this issue.
I'll take a look on it and also make sure we have a test for catching it earlier.
Sorry for the incovinience.

chenrui333:
no worries, thanks for picking this up.

chenrui333:
Any updates?

bmeneguele:
> 
> 
> Any updates?

Not yet, sorry.
Didn't have time to look at it :( but I'll give my word I'll take care of it early next week (Monday).

prarit:
@bmeneguele @chenrui333 :  I think I have a solution.  Will update in a minute.

bmeneguele:
@prarit thanks for the help :) didn't have time to get a closer look at it.
One thing I noticed is that `lab -h` and `lab --help` have different behaviors too: the first requires API auth, while the last doesnt.
But I think it's something for another issue.

prarit:
> @prarit thanks for the help :) didn't have time to get a closer look at it.
[...]

PR #559, opened 2021-01-11
lab: Check that LAB_CORE_TOKEN is set before querying user for a token

prarit:
chenrui333 noted in #558 that the goreleaser test for lab was failing.
The test is a simple execution of the git wrapper capabilities of lab.

The test does:

	git init
	touch haunted
	touch house
	git add haunted house
	git commit -a -m "Initial Commit"
	lab ls-files

and expects the output of the last command to be

	haunted
	house

The test currently fails on the command line with

```
[prarit@prarit dummy (master)]$ LAB_CORE_HOST=foo LAB_CORE_USER=bar LAB_CORE_TOKEN=baz lab  ls-files
Create a token with scope 'api' here: profile/personal_access_tokens
Enter default GitLab token, or leave blank to provide a command to load the token:
```

instead of displaying the list of files.  This occurs because the code
does not check to see if LAB_CORE_TOKEN is set before querying the user
for a token.

Check that LAB_CORE_TOKEN is set before querying the user for the token.

Reported-by: @chenrui333 (in #558)
Cc: Bruno Meneguele <bmeneg@redhat.com>

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -22,7 +22,7 @@ brews:
       system "git", "add", "haunted", "house"
       system "git", "commit", "-a", "-m", "Initial Commit"
       lab_env_config = "LAB_CORE_HOST=foo LAB_CORE_USER=bar LAB_CORE_TOKEN=baz"
-      assert_equal "haunted\nhouse", shell_output("#{lab_env_config} #{bin}/lab ls-files").strip
+      assert_match "haunted\nhouse", shell_output("#{lab_env_config} #{bin}/lab ls-files").strip
 
 scoop:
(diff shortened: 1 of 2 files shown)


prarit:
Closes #558

prarit:
@chenrui333 can you please retest with this change?

prarit:
There may have to be a follow on change in the goreleaser.yml file to switch from "assert_equal" to "assert_contains".

chenrui333 (approved):

chenrui333:
should be good.

bmeneguele (approved):
LGTM!

PR #560, opened 2021-01-13
mr_list: Support search

fmuellner:
Search is as useful for merge requests as for issues, so support
the same option as `lab issue list`.

### DIFF
--- a/cmd/issue_list.go
+++ b/cmd/issue_list.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"log"
 
+	"github.com/pkg/errors"
 	"github.com/rsteube/carapace"
 	"github.com/spf13/cobra"
 	gitlab "github.com/xanzy/go-gitlab"
@@ -12,12 +13,13 @@ import (
 )
 
 var (
-	issueLabels    []string
-	issueMilestone string
-	issueState     string
-	issueSearch    string
-	issueNumRet    int
-	issueAll       bool
+	issueLabels     []string
+	issueMilestone  string
+	issueState      string
+	issueSearch     string
+	issueNumRet     int
+	issueAll        bool
+	issueExactMatch bool
 )
 
 var issueListCmd = &cobra.Command{
@@ -70,6 +72,13 @@ func issueList(args []string) ([]*gitlab.Issue, error) {
 		OrderBy:   gitlab.String("updated_at"),
 	}
 
+	if issueExactMatch {
+		if issueSearch == "" {
+			return nil, errors.New("Exact match requested, but no search terms provided")
+		}
+		issueSearch = "\"" + issueSearch + "\""
+	}
+
 	if issueSearch != "" {
 		opts.Search = &issueSearch
 	}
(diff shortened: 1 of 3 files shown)


bmeneguele:
@fmuellner do you think it would be useful a "--full-match" flag for adding double-quotes in the terms and forcing that "exact" (it still is case insensitive) match?
It would basically be the same as `lab issue search "\"search terms\""`.

fmuellner:
> do you think it would be useful a "--full-match" flag for adding double-quotes in the terms and forcing that "exact" (it still is case insensitive) match?
> It would basically be the same as `lab issue search "\"search terms\""`.

Do you mean to both `lab mr list` and `lab issue list`? IMHO those two commands should be as consistent as possible.

That said, a flag does sound more convenient than escaping quotes on the command line. I'd prefer `--exact-match` over `--full-match` though - the latter suggests something like full-text search (i.e. searching comments as well) to me.

fmuellner:
> do you think it would be useful a "--full-match" flag for adding double-quotes in the terms and forcing that "exact" (it still is case insensitive) match?
> It would basically be the same as `lab issue search "\"search terms\""`.

Do you mean to both `lab mr list` and `lab issue list`?

bmeneguele:
> 
> 
> > do you think it would be useful a "--full-match" flag for adding double-quotes in the terms and forcing that "exact" (it still is case insensitive) match?
> > It would basically be the same as `lab issue search "\"search terms\""`.
> 
> Do you mean to both `lab mr list` and `lab issue list`? IMHO those two commands should be as consistent as possible.
> 

Yes! Sorry. It must be for both :)

> That said, a flag does sound more convenient than escaping quotes on the command line. I'd prefer `--exact-match` over `--full-match` though - the latter suggests something like full-text search (i.e. searching comments as well) to me.

Hmm, although I agree with your concerning with `--full-match` I also think `--exact-match` misleading, since it gives the impression of a case-sensitive search, which is not. What about only `--match` ?

fmuellner:
> What about only `--match` ?

Maybe as a string flag (`lab issue list --match="foo bar"` instead of `lab issue list "foo bar"` - but then what about `lab issue list --match="foo bar" baz`?). For a boolean, the name sounds too generic to me.

> I also think `--exact-match` is somewhat misleading, since it gives the impression of a case-sensitive search

Right, that's a concern.

OTOH, both Google and DuckDuckGo use "exact match/term":
https://support.google.com/websearch/answer/2466433?hl=en
https://help.duckduckgo.com/duckduckgo-help-pages/results/syntax/

And in both cases, the search is case insensitive (in fact, searching for "cats and dogs" turns up "Cats & Dogs" as top result in both for me)

bmeneguele:
> 
> 
> > What about only `--match` ?
> 
> Maybe as a string flag (`lab issue list --match="foo bar"` instead of `lab issue list "foo bar"` - but then what about `lab issue list --match="foo bar" baz`?). For a boolean, the name sounds too generic to me.
> 

Yeah.. using it as a string would be confusing.

> > I also think `--exact-match` is somewhat misleading, since it gives the impression of a case-sensitive search
> 
> Right, that's a concern.
> 
> OTOH, both Google and DuckDuckGo use "exact match/term":
> https://support.google.com/websearch/answer/2466433?hl=en
> https://help.duckduckgo.com/duckduckgo-help-pages/results/syntax/
> 
> And in both cases, the search is case insensitive (in fact, searching for "cats and dogs" turns up "Cats & Dogs" as top result in both for me)

Hmmm. let's use `--exact-match` then. However, if possible, let it clear that the search is case-insensitive in the flag description.

Thanks @fmuellner !

prarit:
Hi, due to no fault of yours the lab test code was broken.  I have fixed the tests in #575 and #573 .  Can you please rebase to see if your build will past testing?  My sincere apologies for the inconvenience.

bmeneguele (approved):
LGTM

PR #567, opened 2021-01-19
Fix the -h lab option

atenart:
The help describes -h as a valid option:

  Flags:
    -h, --help      help for lab

But using it fails with:

  $ lab -h
  unknown option: -h

Fix this.


prarit:
LGTM!  Thanks for the fix @atenart

PR #568, opened 2021-01-19
Internal git fixes

atenart:
Hi,

This PR fixes one error message in internal/git and the UpstreamBranch function which has a too strict expectation on the local remote names.

Thanks!

### DIFF
--- a/internal/git/git.go
+++ b/internal/git/git.go
@@ -147,15 +147,11 @@ func CurrentUpstreamBranch() (string, error) {
 
 // UpstreamBranch returns the upstream of the specified branch
 func UpstreamBranch(branch string) (string, error) {
-	cmd := New("rev-parse", "--abbrev-ref", branch+"@{upstream}")
-	cmd.Stdout = nil
-	cmd.Stderr = nil
-	ref, err := cmd.Output()
+	upstreamBranch, err := gitconfig.Local("branch." + branch + ".merge")
 	if err != nil {
 		return "", errors.Errorf("No upstream for branch '%s'", branch)
 	}
-	upstreamBranch := strings.SplitN(string(ref), "/", 2)[1]
-	return strings.TrimSpace(upstreamBranch), nil
+	return strings.TrimPrefix(upstreamBranch, "refs/heads/"), nil
 }
 
 // PathWithNameSpace returns the owner/repository for the current repo
--- a/internal/git/git.go
+++ b/internal/git/git.go
@@ -349,7 +349,7 @@ func NumberCommits(sha1, sha2 string) int {
 	cmd.Stderr = nil
 	CmdOut, err := cmd.Output()
 	if err != nil {
-		fmt.Printf("There are no commits between %s and %s", sha2, sha2)
+		fmt.Printf("There are no commits between %s and %s\n", sha1, sha2)
 		log.Fatal(err)
 	}
 	numLines := strings.Count(string(CmdOut), "\n")


prarit:
Hi, due to no fault of yours the lab test code was broken.  I have fixed the tests in #575 and #573 .  Can you please rebase to see if your build will past testing?  My sincere apologies for the inconvenience.

atenart:
@prarit Sure, no problem, I'll rebase and update this PR.

internal/git/git.go
```diff
@@ -147,15 +147,12 @@ func CurrentUpstreamBranch() (string, error) {
 
 // UpstreamBranch returns the upstream of the specified branch
 func UpstreamBranch(branch string) (string, error) {
-	cmd := New("rev-parse", "--abbrev-ref", branch+"@{upstream}")
-	cmd.Stdout = nil
-	cmd.Stderr = nil
-	ref, err := cmd.Output()
+	upstreamBranch, err := gitconfig.Local("branch." + branch + ".merge")
```
bmeneguele, line 150: Sorry, but I didn't get the reason for using `branch.<name>.merge` instead of `git rev-parse --abbrev-ref <name>@{usptream}`. At the same time, how this change makes the code "less restrictive" wrt local remote naming?
Could you share a case or workflow where your change has clear advantage?

internal/git/git.go
```diff
@@ -353,7 +350,7 @@ func NumberCommits(sha1, sha2 string) int {
 	cmd.Stderr = nil
 	CmdOut, err := cmd.Output()
 	if err != nil {
-		fmt.Printf("There are no commits between %s and %s", sha2, sha2)
+		fmt.Printf("There are no commits between %s and %s", sha1, sha2)
```
bmeneguele, line 353: Just noticed this also needs a linebreak `\n` at the end of Printf() string.

bmeneguele (changes requested):

internal/git/git.go
atenart: The issue arises when having a remote with at least a '/' in its name. Say we have a remote called `github/username` and a branch named `branch/name`, `git rev-parse --abbrev-ref branch/name@{u}` will return `github/username/branch/name`. 

`UpstreamBranch` only returns the remote branch name, and was splitting the remote name from the branch name based on the first slash, using `strings.SplitN`. That is not valid as remotes can have slashes in their names (and more than one).

We want to fix the logic of removing the remote name from the value returned by `git rev-parse`, however finding this information isn't easy (expect if we loop over all the remote names). But looking at `git rev-parse` documentation, it says the `@{upstream}` value is constructed using `branch.<name>.remote` and `branch.<name>.merge` (meaning we were using `@{upstream}` which was adding the remote name to the branch one, to then remove the remote name part).

Given that, a clean solution is to directly use `branch.<name>.merge` as it is what we want: the remote branch name, without the remote name included. The only thing we need to take care of is that string is prefixed with 'refs/heads/'.

Example:
```
$ git remote
gitlab/username
gitlab/project
$ git rev-parse --abbrev-ref HEAD
my/branch
$ git rev-parse --abbrev-ref @{u}
gitlab/username/my/branch
$ lab mr create gitlab/project
mr_create.go:128: aborting MR, source branch username/my/branch not present on remote gitlab/username.
```

(I just realized we might want to check the string is at least 11 chars long before removing the first 11 chars).

internal/git/git.go
atenart: Good catch!

internal/git/git.go
fmuellner: > (I just realized we might want to check the string is at least 11 chars long before removing the first 11 chars).

Or use `strings.TrimPrefix(upstreamBranch, "refs/heads/")`

internal/git/git.go
```diff
@@ -147,15 +147,12 @@ func CurrentUpstreamBranch() (string, error) {
...
 // UpstreamBranch returns the upstream of the specified branch
 func UpstreamBranch(branch string) (string, error) {
-	cmd := New("rev-parse", "--abbrev-ref", branch+"@{upstream}")
-	cmd.Stdout = nil
-	cmd.Stderr = nil
-	ref, err := cmd.Output()
+	upstreamBranch, err := gitconfig.Local("branch." + branch + ".merge")
 	if err != nil {
-		return "", errors.Errorf("No upstream for branch '%s'", branch)
+		return "", errors.Errorf("No merge for branch '%s'", branch)
```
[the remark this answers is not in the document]
atenart: Yes, I agree. I was going to update it actually when updating the PR (once the discussions will be resolved).

What about "No merge branch for branch '%s'" ? Or we could keep "upstream" if that is more user friendly?

internal/git/git.go
atenart: Right. I always have the impression using those functions when the argument can't be changed at runtime is not ideal. But we're talking about a single call, so I can use `strings.TrimPrefix` if necessary.

bmeneguele: Ah ok. I see the issue now. Thanks for the clarification!!
And I agree with @fmuellner here.. just use `strings.TrimPrefix` here.

internal/git/git.go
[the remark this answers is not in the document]
bmeneguele: IMO keeping the "upstream" word is better.

internal/git/git.go
atenart: All right, I'll use `strings.TrimPrefix` :)

internal/git/git.go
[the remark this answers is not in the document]
atenart: Let's keep "upstream" then.

bmeneguele (approved):
New changes LGTM.
Thanks @atenart

PR #570, opened 2021-01-19
config: Fix "attemping" typo

prarit:
It's attempting not attemping.


prarit:
The failures in the build are not due to anything in this simple patch.  I'm going to merge and then see what the issue is with the builds.

PR #574, opened 2021-01-21
mr approve: Fix unapprove test failures

prarit:
This is failing for some reason.


prarit:
There is still a Test_mrListStateClosed failure that is independent of this fix.

PR #575, opened 2021-01-21
mr_list_test: mrListStateClosed fix

prarit:
mrListStateClosed test is failing with not finding the output "!5 closed
mr".  This occurs because the output contains only the first 20 closed
MRs.  If all the closed MRs were listed the test will pass.

Modify the mrListStateClosed test to list all closed MRs.


prarit:
This fixes the remaining test failure.

PR #576, opened 2021-01-22
Allow configuring a default remote

atenart:
The goal of this PR is to allow the user to configure what should be the default remote. The current logic tries to guess what should be the default remote by looking at the "origin" and "upstream" remotes if defined, and at branches such as "main" and "master". But having those remotes and branches is not always the case, and even so they could hold a wrong value (for lab).

This PR adds the ability to explicitly configure the default remote, using a new `core.default_remote` parameter in the configuration. While it is probably not a good idea to define this globally, the configuration file can also be in the repositories `.git` directories, as well as in the `.git/worktrees/` directories when using Git worktrees. In such cases, having the ability to explicitly set the default remote is quite handy.

This does not break any of the previous cases. When not set, `lab` will strill try to guess what should be the default remote. And commands allowing a remote as a parameter can still do so.

This PR starts with some cosmetic patches and then adds the new `core.default_remote` logic.

Thanks!

### DIFF
--- a/cmd/issue_create.go
+++ b/cmd/issue_create.go
@@ -46,7 +46,7 @@ var issueCreateCmd = &cobra.Command{
 		if err != nil {
 			log.Fatal(err)
 		}
-		remote := forkedFromRemote
+		remote := defaultRemote
 		if len(args) > 0 {
 			ok, err := git.IsRemote(args[0])
 			if err != nil {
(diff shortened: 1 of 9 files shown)


zaquestion:
@atenart this looks good, if we can get the tests/build sorted out happy to build it in. Might just be a matter of bringing in the latest from `master`

atenart:
> @atenart this looks good, if we can get the tests/build sorted out happy to build it in. Might just be a matter of bringing in the latest from `master`

The build actually fails because of an issue in the series: if skipInit is true in main(), the config isn't loaded and we can't access it. I'll push an update with this fixed. Thanks!

bmeneguele (approved):
Changes are looking good to me.
Thanks @atenart

PR #579, opened 2021-02-04
add multi-project CI pipeline support

bmeneguele:
This MR enables `lab` to follow pipelines that are kept in other projects other than the current, where the source code is,
this is called [multi-projects pipeline](https://docs.gitlab.com/ee/ci/multi_project_pipelines.html).

For that, a new persistent flag was added to all `lab ci` commands:
```
  --follow   Follow downstream pipelines in a multi-projects setup
```
e.g.
```
$ lab ci artifacts upstream --follow --merge-request
```

One of the issues here is that we don't have a test infrastructure to properly test it.
I've tested within the company's infrastructure that I work and seems to be fully working.
I'm afraid the only thing we can do is to wait for people to use it.

### DIFF
--- a/cmd/ci.go
+++ b/cmd/ci.go
@@ -4,6 +4,9 @@ import (
 	"github.com/spf13/cobra"
 )
 
+// Hold --follow flag value that is common for all ci command
+var followBridge bool
+
 // ciCmd represents the ci command
 var ciCmd = &cobra.Command{
 	Use:   "ci",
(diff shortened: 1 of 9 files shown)


bmeneguele:
Quick note: the `Bridge` type was added to `go-gitlab` lib in version 0.40.0, however, there were some changes in that code throughout the other releases. Because of that, I updated it to the latest version available (0.43.0).

zaquestion:
Any reason not to make follow the default? Perhaps it doesn't play nicely when downstream pipelines aren't present?

zaquestion (approved):

bmeneguele:
Not really. I just thought that I'm the first one really **requiring** it for my daily workflow, so I thought it would be better to keep everyone else's workflow the same :smile:

PR #581, opened 2021-02-08
Fix label exact match

### LINKED ISSUE #580: Cannot filter MRs using some labels
```
$ lab mr list -l 'Subsystem:net'
2021/02/08 16:21:53 mr_list.go:50: Label 'Subsystem:net' is ambiguous
Error: User authentication failed.  This is likely due to a misconfigured Personal Access Token.  Verify the token or token_load config settings before attempting to authenticate.
```

The error message is misleading (which would be worth fixing, too), it's not a problem with my token. Other queries work fine.

The problem is that in the same project, there is also a Subsystem:netdrv label. It seems that -l matches on any prefix of the label. This makes it impossible to filter on Subsystem:net.

bmeneguele:
This patchset removes a config error message during `mr_list` and make the label ambiguity just fine, instead of returning an error.

Fixes #580

### DIFF
--- a/cmd/label_list.go
+++ b/cmd/label_list.go
@@ -56,6 +56,8 @@ lab label list remote "search term"  # search "remote" for labels with "search t
 }
 
 func MapLabels(rn string, labelTerms []string) ([]string, error) {
+	var ambiguous bool
+
 	labels, err := lab.LabelList(rn)
 	if err != nil {
 		return nil, err
@@ -63,25 +65,40 @@ func MapLabels(rn string, labelTerms []string) ([]string, error) {
 
 	matches := make([]string, len(labelTerms))
 	for i, term := range labelTerms {
+		ambiguous = false
 		lowerTerm := strings.ToLower(term)
 		for _, label := range labels {
 			lowerLabel := strings.ToLower(label.Name)
 
-			// no match, or we already have an exact match
-			if !strings.Contains(lowerLabel, lowerTerm) || matches[i] == lowerLabel {
+			// no match
+			if !strings.Contains(lowerLabel, lowerTerm) {
 				continue
 			}
 
-			// only allow ambiguity for exact matches
+			// check for ambiguity on substring level
 			if matches[i] != "" && lowerTerm != lowerLabel {
-				return nil, errors.Errorf("Label '%s' is ambiguous", term)
+				ambiguous = true
+				continue
 			}
 
 			matches[i] = label.Name
+
+			// exact match
+			// may happen after multiple substring matches
+			if lowerLabel == lowerTerm {
+				ambiguous = false
+				break
+			}
 		}
 
 		if matches[i] == "" {
 			return nil, errors.Errorf("Label '%s' not found", term)
+		}
+
+		// Ambiguous matches should not be returned to avoid
+		// manipulating the wrong label.
+		if ambiguous {
+			return nil, errors.Errorf("Label '%s' has no exact match and is ambiguous\n", term)
 		}
 	}
 
--- a/cmd/mr_list.go
+++ b/cmd/mr_list.go
@@ -9,7 +9,6 @@ import (
 	"github.com/spf13/cobra"
 	gitlab "github.com/xanzy/go-gitlab"
 	"github.com/zaquestion/lab/internal/action"
-	"github.com/zaquestion/lab/internal/config"
 	lab "github.com/zaquestion/lab/internal/gitlab"
 )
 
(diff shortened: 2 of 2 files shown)


prarit:
Even when I use this I can't get "lab mr list -x "Acks::NeedsReview" to work?  Can you try that @bmeneguele ?

bmeneguele:
> Even when I use this I can't get "lab mr list -x "Acks::NeedsReview" to work? Can you try that @bmeneguele ?

The `-x` flag is for searching terms and not labels (basically by putting quotes around the term). For label the `-l` is the only option and the behavior is "exact match" by default. Try changing the `-x` by `-l` ;).

```
$ lab mr list -l "Acks::NeedsReview"
```
works fine here.

prarit (approved):

cmd/label_list.go
```diff
@@ -56,25 +56,29 @@ lab label list remote "search term"  # search "remote" for labels with "search t
...
 			lowerLabel := strings.ToLower(label.Name)
 
-			// no match, or we already have an exact match
-			if !strings.Contains(lowerLabel, lowerTerm) || matches[i] == lowerLabel {
+			// no match
+			if !strings.Contains(lowerLabel, lowerTerm) {
 				continue
 			}
 
 			// only allow ambiguity for exact matches
```
fmuellner, line 78: That comment is no longer valid.

cmd/label_list.go
```diff
@@ -83,6 +87,10 @@ func MapLabels(rn string, labelTerms []string) ([]string, error) {
 		if matches[i] == "" {
 			return nil, errors.Errorf("Label '%s' not found", term)
 		}
+
+		if ambiguous {
+			fmt.Printf("Label '%s' is ambiguous and may be a substring of another label\n", term)
```
fmuellner, line 92: > just warn the user about it and keep the exact match.

That's not what the patch is doing, is it?

If there are substring matches for more than one label, then there is no exact match and the first match is returned. That feels quite arbitrary to me, in particular for actions like create/update, and especially for `label delete`.

cmd/label_list.go
```diff
@@ -56,25 +56,29 @@ lab label list remote "search term"  # search "remote" for labels with "search t
...
 			}
 
 			// only allow ambiguity for exact matches
 			if matches[i] != "" && lowerTerm != lowerLabel {
-				return nil, errors.Errorf("Label '%s' is ambiguous", term)
+				ambiguous = true
+				continue
 			}
 
 			matches[i] = label.Name
```
fmuellner, line 84: Maybe do something like
```
    ambiguous = false;
    if lowerTerm == lowerLabel {
        break;
    }
```
here? That would set `ambiguous` if there are more than one substring matches, but no exact match (which was the intended behavior all along, but I failed to consider multiple substring matches before finding an exact match)

cmd/label_list.go
bmeneguele: Good point. Will remove it.

cmd/label_list.go
bmeneguele: > If there are substring matches for more than one label, then there is no exact match and the first match is returned. 

Oh, of course, I *somehow* missed the `matches[i] != ""` there :/

> That feels quite arbitrary to me, in particular for actions like create/update, and especially for label delete

Not sure I follow your thought here.
The idea of returning the first "sub match" is the "arbitrary" part? If so, I completely agree!
Will fix it in v2.

cmd/label_list.go
bmeneguele: I was thinking a bit more about the `ambiguous` state here: is it even worth informing the user if we're keeping exact match only behavior?

cmd/label_list.go
fmuellner: > The idea of returning the first "sub match" is the "arbitrary" part?

Yes. I'm thinking of cases like `lab label delete test` to delete a "test label" label, and end up deleting a "User Testing" label instead.

cmd/label_list.go
fmuellner: > is it even worth informing the user if we're keeping exact match only behavior?

If there is an exact match, then it's not useful in my opinion.

But I'd still like to allow non-exact matches when a term can be matched to a single label (for selfish reasons, because gitlab.gnome.org's labels aren't very CLI friendly).

bmeneguele: >But I'd still like to allow non-exact matches when a term can be matched to a single label (for selfish reasons, because gitlab.gnome.org's labels aren't very CLI friendly).

ahh ok! I remember you sharing the link to them in some PR :).
I'm going to rework this patch.

Thanks for all the feedback @fmuellner

bmeneguele: Should be fixed in the new patch version

fmuellner: Thanks!

Looks good to me at a quick glance.

Issue #582, opened 2021-02-10
approve and unapprove message failures can be better


prarit:
When I attempt to approve or unapprove on a Project that I don't have approve permissions for, I get the following error:

2021/02/10 10:58:17 mr_approve.go:56: POST https://gitlab.com/api/v4/projects/XXXX/merge_requests/208/approve: 403 {message: 403 Forbidden}

It would be good to provide some better error messaging.

And, yes, I already have a patch :)

bmeneguele:
Closing issue based on the PR already merged.

PR #583, opened 2021-02-10
582

prarit:
mr approve/unapprove: Provide better error messages

When I attempt to approve or unapprove on a Project that I don't have approve permissions for, I get the following error:
   
`      2021/02/10 10:58:17 mr_approve.go:56: POST https://gitlab.com/api/v4/projects/XXXX/merge_requests/208/approve: 403 {message: 403 Forbidden}`
    
Provide a better error message in this case.

### DIFF
--- a/internal/gitlab/gitlab.go
+++ b/internal/gitlab/gitlab.go
@@ -30,6 +30,8 @@ var (
 	ErrProjectNotFound = errors.New("gitlab project not found, verify you have access to the requested resource")
 	// ErrGroupNotFound is returned when a GitLab group cannot be found.
 	ErrGroupNotFound = errors.New("gitlab group not found")
+	// ErrStatusForbidden is returned when attempting to access a GitLab project with insufficient permissions
+	ErrStatusForbidden = errors.New("Insufficient permissions for gitlab project")
 )
 
 var (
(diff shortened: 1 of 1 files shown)


prarit:
Uh something went incredibly wrong here.  @bmeneguele not sure what happened.  I think you'll have to revert my commits and repush yours.

prarit:
Looks like I somehow pushed a polluted branch.  My bad.  Fixed now.

bmeneguele:
Ok, this build issue has happened to every PR since Feb 5th, and the reason for that is:
```
"artifacts_expire_at": "2021-02-05T20:28:01.372Z",
```
CI artifacts seem to have an expiration date. I'm checking how it can be turned off or extended or worked around for our testing.

bmeneguele:
> I'm checking how it can be turned off or extended or worked around for our testing.

It seems we have something to try:
```
POST /projects/:id/jobs/:job_id/retry
POST /projects/:id/jobs/:job_id/artifacts/keep
```

cmd/mr_approve.go
```diff
@@ -52,6 +54,11 @@ var mrApproveCmd = &cobra.Command{
 
 		err = lab.MRApprove(p.ID, int(id))
 		if err != nil {
+			if strings.Contains(err.Error(), "{message: 403 Forbidden}") {
```
bmeneguele, line 57: Instead of comparing to the error string, wouldn't it be better to check the "resp" return value in `internal/gitlab/gitlab.go:MRApprove()` returned by `lab.MergeRequestApprovals.ApproveMergeRequest()` and create a new error type as it's already done for both `ErrProjectNotFound` and `ErrGroupNotFound` ?

bmeneguele (commented):
The bellow comment is valid for both MRApprove and MRUnapprove.

cmd/mr_approve.go
bmeneguele: Additionally, `http.StatusForbidden` could be used in comparison with the `resp` return value.

prarit: Oh, I had no idea I could do it that way.  Let me look into it!

prarit: Updated :)  Thanks for the suggestion @bmeneguele !

bmeneguele: It seems you dropped the `cmd/mr_{un}approve.go` changes, the ones actually checking the error and printing better logs :)

bmeneguele (approved):
LGTM :)

bmeneguele:
Fixes #582

PR #584, opened 2021-02-10
mr_create: Evaluate label/milstone flags before description

fmuellner:
Both options can fail, for example when the referenced label/milestone
doesn't exist. It is better to fail early in that case, and not wait
until the user composed a MR description in the editor (when none was
passed on the command line).


bmeneguele (approved):
Ha! Just submitted a PR changing the place where "title" emptiness check happens right this very same function. :)

LGTM.

PR #588, opened 2021-02-11
mr_create: check for empty msg before acting

### LINKED ISSUE #587: panic: runtime error: slice bounds out of range [:4] with length 0
When one create a MR passing `--cover-letter` the text editor is opened to allow the one to write a description message.
When the file is saved and closed no content it prints the message: `2021/02/11 10:32:35 mr_create.go:243: aborting MR due to empty MR msg`, which is fine and correct.

However, if the user also passes `--draft` and save&close and empty file:
```
panic: runtime error: slice bounds out of range [:4] with length 0

goroutine 1 [running]:
github.com/zaquestion/lab/cmd.runMRCreate(0x18c9a60, 0xc0005f6600, 0x0, 0x3)
        /home/bmeneg/git/lab/cmd/mr_create.go:203 +0x1da5
github.com/spf13/cobra.(*Command).execute(0x18c9a60, 0xc0005f65d0, 0x3, 0x3, 0x18c9a60, 0xc0005f65d0)
        /home/bmeneg/go/pkg/mod/github.com/spf13/cobra@v1.1.1/command.go:854 +0x2c2
github.com/spf13/cobra.(*Command).ExecuteC(0x18caf60, 0xc000032240, 0x6, 0x6)
        /home/bmeneg/go/pkg/mod/github.com/spf13/cobra@v1.1.1/command.go:958 +0x375
github.com/spf13/cobra.(*Command).Execute(...)
        /home/bmeneg/go/pkg/mod/github.com/spf13/cobra@v1.1.1/command.go:895
github.com/zaquestion/lab/cmd.Execute()
        /home/bmeneg/git/lab/cmd/root.go:241 +0x188
main.main()
        /home/bmeneg/git/lab/main.go:28 +0x93
```

I'm going to take a look at that later.

bmeneguele:
Fix panic during a `lab mr create -c --draft` with an empty cover-letter.

Fixes #587

### DIFF
--- a/cmd/mr_create.go
+++ b/cmd/mr_create.go
@@ -229,10 +229,10 @@ func runMRCreate(cmd *cobra.Command, args []string) {
 
 	draft, _ := cmd.Flags().GetBool("draft")
 	if draft {
-		isWIP := strings.EqualFold(title[0:4], "wip:")
-		isDraft := strings.EqualFold(title[0:6], "draft:") ||
-			strings.EqualFold(title[0:7], "[draft]") ||
-			strings.EqualFold(title[0:7], "(draft)")
+		isWIP := hasPrefix(title, "wip:")
+		isDraft := hasPrefix(title, "draft:") ||
+			hasPrefix(title, "[draft]") ||
+			hasPrefix(title, "(draft)")
 
 		if !isWIP && !isDraft {
 			title = "Draft: " + title
--- a/cmd/mr_create.go
+++ b/cmd/mr_create.go
@@ -218,6 +218,10 @@ func runMRCreate(cmd *cobra.Command, args []string) {
 		}
 	}
 
+	if title == "" {
+		log.Fatal("aborting MR due to empty MR msg")
+	}
+
 	linebreak, _ := cmd.Flags().GetBool("force-linebreak")
 	if linebreak {
 		body = textToMarkdown(body)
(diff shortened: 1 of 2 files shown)


cmd/mr_create.go
```diff
@@ -193,6 +193,10 @@ func runMRCreate(cmd *cobra.Command, args []string) {
 		}
 	}
 
+	if title == "" {
+		log.Fatal("aborting MR due to empty MR msg")
+	}
+
```
fmuellner, line 199: This makes sense to me.

However:

```
		isWIP := strings.EqualFold(title[0:4], "wip:")
```

this probably still panics if title is not empty, but shorter than 4 characters?

bmeneguele: Should be fixed with the new patch.
Thanks Florian!

cmd/mr_create.go
```diff
@@ -204,10 +204,16 @@ func runMRCreate(cmd *cobra.Command, args []string) {
...
-		isDraft := strings.EqualFold(title[0:6], "draft:") ||
-			strings.EqualFold(title[0:7], "[draft]") ||
-			strings.EqualFold(title[0:7], "(draft)")
+		var isWIP, isDraft bool
+
+		// Regex could be used, but we have only a limited set of words for
+		// draft mode, so use hardcoded limits
+		if len(title) >= 4 {
+			isWIP = strings.EqualFold(title[0:4], "wip:")
+			isDraft = strings.EqualFold(title[0:6], "draft:") ||
```
fmuellner, line 213: You moved the crash one line :wink: 

How about adding a small helper function like
```
func hasPrefix(str, prefix string) bool {
    if len(str) < len(prefix) {
        return false;
    }
    return string.EqualFold(str[0:len(prefix)], prefix)
}
```

bmeneguele: Wow! Goosh! How could I just miss/mix the minimum vs maximum size!
Well.. today has been a really busy day.. should avoid working in many different project's codes at the same time heheh :/.

I was tempted to use the `strings.hasPrefix()` for that... but it's case sensitive. Will follow your suggestion.
Thanks for the help Florian!

fmuellner: Looks good to me at a glance, but it has been a busy day here as well, so take that with a grain of salt.

bmeneguele: :smile: tomorrow I'll recheck the whole thing again.

bmeneguele:
Had to perform a simple rebase on top of master.

PR #593, opened 2021-02-17
mr_list: fix help message wrt MR state

bmeneguele:
Add `all` to the `lab mr list --help` command under `-s, --state string` paramenter.
With that the user don't need to guess this option is available.


fmuellner:
I guess the same should be done for `issue list`.

PR #595, opened 2021-02-21
mr_create: Add number of commits to create metadata

prarit:
First time GitLab users can make mistakes that result in a large merge
request being generated.  For example, new users may get the target branch
wrong, or push a branch that hasn't been updated, etc.  These errors always
result in way more commits than intended being pushed.

After a push to one of their hosted repositories, GitLab provides a URL
to a webpage that allows users to change the target branch and look at the
commits that are going to be pushed.  One convenient value on that page is
the number of commits that are being pushed.  In order to prevent users
from making simple mistakes, like the above, it would be good to also
provide that information the default 'lab mr create' message.

Provide the number of commits in 'lab mr create'.

### DIFF
--- a/cmd/mr_create.go
+++ b/cmd/mr_create.go
@@ -282,7 +282,7 @@ func mrText(base, head, sourceRemote, targetRemote string, coverLetterFormat boo
 	const tmpl = `{{if .InitMsg}}{{.InitMsg}}{{end}}
 
 {{if .Tmpl}}{{.Tmpl}}{{end}}
-{{.CommentChar}} Requesting a merge into {{.Base}} from {{.Head}}
+{{.CommentChar}} Requesting a merge into {{.Base}} from {{.Head}} ({{.NumCommits}} commits)
 {{.CommentChar}}
 {{.CommentChar}} Write a message for this merge request. The first block
 {{.CommentChar}} of text is the title and the rest is the description.{{if .CommitLogs}}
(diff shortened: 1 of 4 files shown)


prarit:
@bmeneguele , this results in the default 'lab mr create' message changing from

`# Requesting a merge into upstream:8.3 from prarit:83_test`

to, for example,

`# Requesting a merge into upstream:8.3 from prarit:83_test (10 commits)`

Note that I am keeping it simple and using 'commits'.  Yeah, I could have done a singular/plural thing there but will implement that later on when I get sufficiently annoyed at it :)

bmeneguele:
> Note that I am keeping it simple and using 'commits'. Yeah, I could have done a singular/plural thing there but will implement that later on when I get sufficiently annoyed at it :)

:smile: no problem. 

@prarit but I think you forgot to update the test using the template you modified (added the commit number ;).

bmeneguele:
@prarit actually the build test is failing because both
https://github.com/zaquestion/lab/blob/f3c8fb228452ac5a1e246517590c655c6245a37e/cmd/mr_create_test.go#L20
https://github.com/zaquestion/lab/blob/f3c8fb228452ac5a1e246517590c655c6245a37e/cmd/mr_create_test.go#L40
It's required the `(%d commits)` at the end of the line, which was added by your patch :)

bmeneguele (approved):

PR #602, opened 2021-02-25
fix project lookup during fork

bmeneguele:
Check if the project the user wants to fork matches the project that another fork, with the same name, was based on.


bmeneguele:
Oops.. the commit " Completely drop hub support " was meant to another PR in the near future.
For now, it can be ignored :)

zaquestion (approved):

PR #603, opened 2021-02-25
Add subscribe/unsubscribe subcommands

fmuellner:
For projects where one is only tangentially involved, it sometimes
makes sense to subscribe to an individual issue or merge request
rather than the project as a whole.
    
Support that with corresponding subcommands.


bmeneguele (approved):
Thanks @fmuellner . It was my next item in my TODO list :D.
LGTM.

PR #608, opened 2021-03-02
allow user to update target branch while editing an MR

bmeneguele:
Add the `--target-branch` (or `-t`) flag to the mr_edit command, allowing the user to update the target branch for that specific MR through `lab`.

### DIFF
--- a/cmd/label_list.go
+++ b/cmd/label_list.go
@@ -56,50 +56,19 @@ lab label list remote "search term"  # search "remote" for labels with "search t
 }
 
 func MapLabels(rn string, labelTerms []string) ([]string, error) {
-	var ambiguous bool
-
 	labels, err := lab.LabelList(rn)
 	if err != nil {
 		return nil, err
 	}
 
-	matches := make([]string, len(labelTerms))
-	for i, term := range labelTerms {
-		ambiguous = false
-		lowerTerm := strings.ToLower(term)
-		for _, label := range labels {
-			lowerLabel := strings.ToLower(label.Name)
+	labelNames := make([]string, len(labels))
+	for _, label := range labels {
+		labelNames = append(labelNames, label.Name)
+	}
 
-			// no match
-			if !strings.Contains(lowerLabel, lowerTerm) {
-				continue
-			}
-
-			// check for ambiguity on substring level
-			if matches[i] != "" && lowerTerm != lowerLabel {
-				ambiguous = true
-				continue
-			}
-
-			matches[i] = label.Name
-
-			// exact match
-			// may happen after multiple substring matches
-			if lowerLabel == lowerTerm {
-				ambiguous = false
-				break
-			}
-		}
-
-		if matches[i] == "" {
-			return nil, errors.Errorf("Label '%s' not found", term)
-		}
-
-		// Ambiguous matches should not be returned to avoid
-		// manipulating the wrong label.
-		if ambiguous {
-			return nil, errors.Errorf("Label '%s' has no exact match and is ambiguous\n", term)
-		}
+	matches, err := matchTerms(labelTerms, labelNames)
+	if err != nil {
+		return nil, errors.Errorf("Label %s\n", err.Error())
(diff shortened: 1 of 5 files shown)


internal/gitlab/gitlab.go
```diff
@@ -874,6 +874,49 @@ func LabelDelete(project, name string) error {
...
+		return branches[0], nil
+	case 0:
+		return nil, errors.Errorf("Branch '%s' not found", name)
+	default:
+		return nil, errors.Errorf("Branch '%s' is ambiguous", name)
+	}
+}
+
+// BranchList get all branches from the project that somehow matches the
+// name requested
```
fmuellner, line 895:
```suggestion
// requested options
```

maybe? It's used for matching on a name in BranchGet, but the function itself is more generic.

internal/gitlab/gitlab.go
```diff
@@ -874,6 +874,49 @@ func LabelDelete(project, name string) error {
...
+	}
+
+	branches, _ := BranchList(project, opts)
+	switch len(branches) {
+	case 1:
+		return branches[0], nil
+	case 0:
+		return nil, errors.Errorf("Branch '%s' not found", name)
+	default:
+		return nil, errors.Errorf("Branch '%s' is ambiguous", name)
```
fmuellner, line 890: I think this needs something like the exact-match handling we do for label names. Otherwise it's impossible to set the target branch to `mainline` if a `mainline-next` branch exists as well, no?

internal/gitlab/gitlab.go
bmeneguele: Yep, well pointed.

internal/gitlab/gitlab.go
bmeneguele: Hmmm.. indeed! Good catch. Will add the exact-match handling.

bmeneguele:
@fmuellner, I moved the match logic, including the ambiguity check, to a generic function in cmd/util.go, thus I used it in the target-branch checking as well.

cmd/util.go
```diff
@@ -505,6 +505,53 @@ func hasPrefix(str, prefix string) bool {
...
+				break
+			}
+		}
+
+		if matches[i] == "" {
+			return nil, errors.Errorf("'%s' not found", sTerm)
+		}
+
+		// Ambiguous matches should not be returned to avoid
+		// manipulating the wrong label.
```
fmuellner, line 546: It's no longer just about labels. "item" maybe?

cmd/label_list.go
```diff
@@ -68,7 +68,7 @@ func MapLabels(rn string, labelTerms []string) ([]string, error) {
 
 	matches, err := matchTerms(labelTerms, labelNames)
 	if err != nil {
-		return nil, errors.Errorf("Label %s", err.Error())
+		return nil, errors.Errorf("Label %s\n", err.Error())
 	}
```
fmuellner, line 72: Should be in the previous commit

cmd/util.go
```diff
@@ -545,7 +545,7 @@ func matchTerms(searchTerms, existentTerms []string) ([]string, error) {
 		// Ambiguous matches should not be returned to avoid
 		// manipulating the wrong label.
 		if ambiguous {
-			return nil, errors.Errorf("'%s' has no exact match and is ambiguous\n", sTerm)
+			return nil, errors.Errorf("'%s' has no exact match and is ambiguous", sTerm)
 		}
```
fmuellner, line 549: This as well.

fmuellner (commented):
Looks good to me now except for a couple of nitpicks :-)

cmd/util.go
bmeneguele: Of course.

cmd/label_list.go
bmeneguele: darn.. yes. of course. thanks!

fmuellner (approved):

PR #612, opened 2021-03-03
Make installer script POSIX compliant

### LINKED ISSUE #339: Install script is not Posix
Two problems here:

- The script is dependent to bash. Make it compatible with sh would simplify the installation process on tiny os. Useful for Docker application
- The `--only` grep option does not exist on all grep binaries.

profclems:
Make installer script POSIX compliant by removing bashisms and make it compatible with sh

Resolves #339


prarit (commented):
This looks good to me.  Thanks for the changes.

prarit (approved):

PR #619, opened 2021-03-03
add `author` filter and remove `mine` from lists

### LINKED ISSUE #617: What does it mean for an MR to be "assigned" to someone?
At first glance I expected "lab mr list --mine" to show the pending MRs that I have submitted.

But the help text does say that --mine will show the MRs that are "assigned" to me. Does that mean MRs for which I'm a required reviewer?

If that's the case, it would be useful to have some listing option to show me the pending MRs that I (or a specified user) have submitted.

bmeneguele:
The option `--mine` in `lab mr list`, both to me and others, is kinda misleading since it gives the impression of "I'm the author of that". This patchset removes it from mr_list and adds the `--author` for both mr_list and issue_list.

Fixes #617 .

### DIFF
--- a/cmd/mr_list.go
+++ b/cmd/mr_list.go
@@ -21,13 +21,15 @@ var (
 	mrNumRet       string
 	mrAll          bool
 	mrMine         bool
+	mrAuthor       string
+	mrAuthorID     *int
 	mrDraft        bool
 	mrReady        bool
 	mrConflicts    bool
 	mrNoConflicts  bool
 	mrExactMatch   bool
-	assigneeID     *int
 	mrAssignee     string
+	mrAssigneeID   *int
 	order          string
 	sortedBy       string
 )
@@ -75,18 +77,28 @@ func mrList(args []string) ([]*gitlab.MergeRequest, error) {
 		num = -1
 	}
 
+	// gitlab lib still doesn't have search by assignee and author username
+	// for merge requests, because of that we need to get the ID for both.
 	if mrAssignee != "" {
-		_assigneeID, err := lab.UserIDByUserName(mrAssignee)
+		assigneeID, err := lab.UserIDByUserName(mrAssignee)
 		if err != nil {
 			log.Fatal(err)
 		}
-		assigneeID = &_assigneeID
+		mrAssigneeID = &assigneeID
 	} else if mrMine {
-		_assigneeID, err := lab.UserID()
+		assigneeID, err := lab.UserID()
 		if err != nil {
 			log.Fatal(err)
 		}
-		assigneeID = &_assigneeID
+		mrAssigneeID = &assigneeID
+	}
+
+	if mrAuthor != "" {
+		authorID, err := lab.UserIDByUserName(mrAuthor)
+		if err != nil {
+			log.Fatal(err)
+		}
+		mrAuthorID = &authorID
 	}
 
 	if mrMilestone != "" {
@@ -114,7 +126,8 @@ func mrList(args []string) ([]*gitlab.MergeRequest, error) {
 		Milestone:              &mrMilestone,
 		OrderBy:                orderBy,
 		Sort:                   sort,
-		AssigneeID:             assigneeID,
+		AuthorID:               mrAuthorID,
+		AssigneeID:             mrAssigneeID,
 		WithMergeStatusRecheck: gitlab.Bool(mrCheckConflicts),
 	}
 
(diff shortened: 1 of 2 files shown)


prarit:
@zaquestion , I'm fine and in favor of this change as 'mine' is confusing.  I just dunno what your policy is for backwards compatibility.  I could see you wanting @bmeneguele to keep the --mine option if you're worried about that sort of thing.

prarit (approved):

bmeneguele:
> @zaquestion , I'm fine and in favor of this change as 'mine' is confusing. I just dunno what your policy is for backwards compatibility. I could see you wanting @bmeneguele to keep the --mine option if you're worried about that sort of thing.

Yes, that's precisely my fear: backwards compatibility. @zaquestion if you prefer to keep it for the sake of compatibility I'm fine with it too, no worries, I would only add `--author` then. However, in my defense: one of the ideas of the major version being less then `1` (`0.20.0`) "means" (to me, at least, :smile:) we don't apply any compatibility level in the project. But it's your call @zaquestion.

fmuellner:
> if you prefer to keep it for the sake of compatibility I'm fine with it too, no worries, I would only add `--author` then.

Looking at https://pkg.go.dev/github.com/spf13/pflag#Flag, it's possible to mark `--mine` as deprecated in favor of `--author` and hide it from `--help`.

cmd/mr_list.go
```diff
@@ -74,18 +75,22 @@ func mrList(args []string) ([]*gitlab.MergeRequest, error) {
...
 		if err != nil {
 			log.Fatal(err)
 		}
-		assigneeID = &_assigneeID
-	} else if mrMine {
-		_assigneeID, err := lab.UserID()
+		mrAssigneeID = &assigneeID
+	}
+
+	if mrAuthor != "" {
```
fmuellner, line 88: I was thinking about allowing `--author` without argument as a fancy shorthand for "authored-by-me", but it's probably a bad idea given that `--assignee=""` means "not assigned to anybody".

prarit:
> > if you prefer to keep it for the sake of compatibility I'm fine with it too, no worries, I would only add `--author` then.
> 
> Looking at https://pkg.go.dev/github.com/spf13/pflag#Flag, it's possible to mark `--mine` as deprecated in favor of `--author` and hide it from `--help`.

Yeah, I was thinking exactly the same thing.  Perhaps that just enough to do it and move on.

cmd/mr_list.go
bmeneguele: yea. I think it would be again confusing.
I thought about creating `--authored-by-me` and `--assigned-to-me` or something like that later.. but at the same time it isn't that important considering both are just shorthand for something possible with `--author` and `--assignee`.

bmeneguele:
> > if you prefer to keep it for the sake of compatibility I'm fine with it too, no worries, I would only add `--author` then.
> 
> Looking at https://pkg.go.dev/github.com/spf13/pflag#Flag, it's possible to mark `--mine` as deprecated in favor of `--author` and hide it from `--help`.

Hmmm, thanks for pointing me to that! Didn't know about that.
It for sure is more suitable for this case.

bmeneguele:
Code is now rebased on master and `--mine` is marked as deprecated now.

cmd/mr_list.go
```diff
@@ -172,6 +185,8 @@ func init() {
 		&mrMilestone, "milestone", "", "list only MRs for the given milestone")
 	listCmd.Flags().BoolVarP(&mrAll, "all", "a", false, "list all MRs on the project")
 	listCmd.Flags().BoolVarP(&mrMine, "mine", "m", false, "list only MRs assigned to me")
+	listCmd.Flags().MarkDeprecated("mine", "please --assignee instead")
```
fmuellner, line 188:
```suggestion
	listCmd.Flags().MarkDeprecated("mine", "please use --assignee instead")
```

fmuellner (commented):
One nit, otherwise lgtm

cmd/mr_list.go
prarit: And one other thing :)  don't use 'please'.  I know it sounds rude not to but it's something that you're not supposed to do.  I'm definitely guilty of writing 'polite' instructions too :)

bmeneguele: Yea, you're right. I'm going to drop the `please` and add the `use` that I forgot.

fmuellner (approved):

PR #620, opened 2021-03-03
added label completion

rsteube:
Works fine with the default remote but i am not too sure about the remote arg parsing with `parseArgsRemoteAndProject` (works fine with a fork so i assume it's correct).
[Cache](https://rsteube.github.io/carapace/carapace/cache.html) is written to `/tmp/carapace/{user}/lab/`

### DIFF
--- a/cmd/issue_create.go
+++ b/cmd/issue_create.go
@@ -169,6 +169,13 @@ func init() {
 	issueCmd.AddCommand(issueCreateCmd)
 
 	carapace.Gen(issueCreateCmd).FlagCompletion(carapace.ActionMap{
+		"label": carapace.ActionMultiParts(",", func(c carapace.Context) carapace.Action {
+			if project, _, err := parseArgsRemoteAndProject(c.Args); err != nil {
+				return carapace.ActionMessage(err.Error())
+			} else {
+				return action.Labels(project).Invoke(c).Filter(c.Parts).ToA()
+			}
+		}),
 		"milestone": carapace.ActionCallback(func(c carapace.Context) carapace.Action {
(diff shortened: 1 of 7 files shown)


bmeneguele:
Yeah, it seems to be working fine with the tests I did.
But just a question: is it possible to pass a "custom" pager to `carapace`?

I tested in a fork that uses a lot of labels and `more` popped up as the pager, but by default, I'm used to using what I've configured in git config `core.pager`, like done by `NewPager()` when a lab command output requires a pager.

rsteube:
Hmm..isn't that the configuration of your shell? `more` sounds sounds like bash which as i've read recently isn't even the `more` pager but rather sth. bash provides (which can't scroll back up).

bmeneguele:
Not really. I can set an environment variable to use the pager I want, but `lab` also check for the `core.pager` config from git.
But, at the same time, I don't think it's indeed required.

Could you rebase this MR code so it can be merged? :)

rsteube:
No i mean the paging you see comes from the bash completion engine and not from lab. The output of lab can indeed piped in a pager of your choice but the completion carapace provides is handled by the shell.

Will do a rebase.

bmeneguele:
> No i mean the paging you see comes from the bash completion engine and not from lab. The output of lab can indeed piped in a pager of your choice but the completion carapace provides is handled by the shell.
> https://unix.stackexchange.com/questions/158269/can-we-use-something-better-than-more-for-tab-completion

Ahhh.. I see what you meant now! Ok, that's a piece of interesting information.
Thanks!!

bmeneguele (approved):

PR #621, opened 2021-03-04
lab: Clean up IntVarP usage

prarit:
A user noticed an issue in which 'lab mr --list' and 'lab mr list'
returned lists of 10 and 20 entries respectively.  Both effectively call
the same code in the mr_list file and the only difference is that
LabPersistentPreRun() is run on 'lab mr' and 'lab mr list' respectively.

It appears that IntVarP is not handled properly by viper (there are a few
reports of other issues with IntVarP online) and the suggestion is to
use StringVarP in place of IntVarP.

The IntVarP value is used as the number of entries returned per page
for the list commands of MRs, Issues, Projects, and Snippets.  In all of
these cases the user modified value was not being honoured in the
code so fixes for these are also included.

### DIFF
--- a/cmd/issue_list.go
+++ b/cmd/issue_list.go
@@ -3,6 +3,7 @@ package cmd
 import (
 	"fmt"
 	"log"
+	"strconv"
 
 	"github.com/pkg/errors"
 	"github.com/rsteube/carapace"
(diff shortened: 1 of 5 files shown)


bmeneguele (approved):
Ok, that was indeed a really nice catch.
I've tested it locally and seems to be working as expected!
Thanks @prarit

PR #623, opened 2021-03-04
lab: Remove git version from version command

prarit:
The 'lab version' command prints out the git version.  For example,

[prarit@prarit ~]$ lab version
git version 2.29.2
lab version master

Remove the git version output and move the option into the root file.


bmeneguele (approved):
Well, I don't think `git` version here is indeed something too important, even because we don't have "feature" mapping between lab version and git version.
With that in mind, I'm fine with the change.

PR #629, opened 2021-03-09
mr,issue,snippet: make shorthand flags deprecated

bmeneguele:
Deprecate shorthand flags directly in the commands (mr, issue and snippet).

These shorthands don't really work as expected: the flags from the subcommands are not available for them, so it has been basically a shorthand for the default behavior of a subcommand: `lab mr list` == `lab mr --list`, and that's it. IMHO, the subcommand is easier to type than the shorthand, so I don't think it makes sense to keep them around.


prarit (approved):
LGTM

PR #634, opened 2021-03-15
mr checkout: Add --force option

prarit:
Users have reported a minor annoyance with 'lab mr checkout':

lab mr checkout 3
lab mr checkout 3

and is returned a "ERROR: mr 3 branch branchname already
exists."

In most cases that I can think of the person checking out the MR
wants the latest version of the the code in the MR, not old code.  It
would be convenient to have a --force option that would delete and
overwrite the existing MR branch.

Add 'lab mr checkout --force' option that overwrites an existing MR branch
checkout.


bmeneguele (approved):
LGTM

PR #637, opened 2021-03-16
mr_create: Check for zero commits in a Merge Request

prarit:
'lab mr create' can incorrectly report the number of commits in a merge reuqest.  This is because the code incorrectly looks at the local source branch rather than the remote source branch when checking the number of commits.  Additionally, the 'lab mr create' also does not error if there are zero commits found between the remote source and remote target branches. 

Fix the number of commits reported by examining the remote source branch and return an error if the number of commits is zero.  

### DIFF
--- a/cmd/mr_create.go
+++ b/cmd/mr_create.go
@@ -115,36 +115,34 @@ func runMRCreate(cmd *cobra.Command, args []string) {
 	if err != nil {
 		log.Fatal(err)
 	}
-	branch, err := git.CurrentBranch()
+
+	localBranch, err := git.CurrentBranch()
 	if err != nil {
 		log.Fatal(err)
 	}
 
-	sourceRemote := determineSourceRemote(branch)
+	sourceRemote := determineSourceRemote(localBranch)
 	sourceProjectName, err := git.PathWithNameSpace(sourceRemote)
 	if err != nil {
 		log.Fatal(err)
 	}
 
-	remoteBranch, err := git.CurrentUpstreamBranch()
-	if remoteBranch == "" {
-		// Fall back to local branch
-		remoteBranch, err = git.CurrentBranch()
-	}
-
-	if err != nil {
-		log.Fatal(err)
+	// We want the pushed branch name
+	sourceBranch, _ := git.UpstreamBranch(localBranch)
+	if sourceBranch == "" {
+		// Fall back to local branch name
+		sourceBranch = localBranch
 	}
 
 	p, err := lab.FindProject(sourceProjectName)
 	if err != nil {
 		log.Fatal(err)
 	}
-	if _, err := lab.GetCommit(p.ID, remoteBranch); err != nil {
+	if _, err := lab.GetCommit(p.ID, sourceBranch); err != nil {
 		err = errors.Wrapf(
 			err,
 			"aborting MR, source branch %s not present on remote %s. did you forget to push?",
-			remoteBranch, sourceRemote)
+			sourceBranch, sourceRemote)
 		log.Fatal(err)
 	}
 
@@ -172,7 +170,7 @@ func runMRCreate(cmd *cobra.Command, args []string) {
 			err = errors.Wrapf(
 				err,
 				"aborting MR, %s:%s is not a valid target. Did you forget to push %s to %s?",
-				targetRemote, branch, branch, targetRemote)
+				targetRemote, localBranch, localBranch, targetRemote)
 			log.Fatal(err)
 		}
 	}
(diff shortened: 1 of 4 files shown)


bmeneguele (approved):
Changes are looking good to me.
Only thing is to correct the test code for the new git.Log() return values:  
`internal/git/git_test.go:76:18: assignment mismatch: 3 variables but Log returns 2 values`

prarit:
Yep.  I also have to fix the code to look at the remote source branch instead of the local source branch when calculating the change log and number of commits.  Hopefully this morning :)

bmeneguele:
Ah yes! well observed.

prarit:
Just updating the text in the commit descriptions.  Remaining work is to provide a test for 'lab mr create --source'.

bmeneguele:
@prarit, both `git.CurrentBranch` and `git.CurrentUpstreamBranch()` must take in consideration the `remote:branch` the user passed through args now.

https://github.com/zaquestion/lab/blob/a897827b36989c87b88020879aa50899b74e0c81/cmd/mr_create.go#L129-L133

The above snippet will check `remoteBranch` for its existence in the `sourceRemote`, which will fail if the user is checked out in a branch different from the one he passed through the args. For instance:
```
$ git remote
origin
upstream
$ git branch -vv
* 8.3              fc08cc6b47fb [upstream/8.3] test
genlog        07d8b29d4e25 [origin/genlog] [redhat] fix genlog.py for patches with single line changelog
main           dabd49203c52 [upstream/main] Merge branch '1872412' into 'main'
$ lab mr create --source origin:genlog-test upstream main
2021/03/17 12:01:47 mr_create.go:169: aborting MR, source branch 8.3 not present on remote origin. did you forget to push?: GET https://gitlab.com/api/v4/projects/21035236/repository/commits/8.3: 404 {message: 404 Commit Not Found}
```
I was checked out in branch `8.3`.

Another thing, but just cosmetics: could you also add a newline to the end of the messages being printed with `log.Wrapf()`? :D.

prarit:
>The above snippet will check remoteBranch for its existence in the sourceRemote, which will fail if the user is checked out in a branch different from the one he passed through the args. For instance:

Oh good catch.  And that becomes a good test case too!

prarit:
I am removing the --source option from this PR and will push it separately.  That way at least lab is working correctly and doesn't suffer from the brokenness I introduced :)

cmd/mr_create.go
```diff
@@ -285,17 +285,18 @@ func mrText(base, head, sourceRemote, targetRemote string, coverLetterFormat boo
 		err       error
 	)
 	remoteBase := fmt.Sprintf("%s/%s", targetRemote, base)
+	targetBase := fmt.Sprintf("%s/%s", sourceRemote, head)
```
bmeneguele, line 288: May I kindly suggest that we change these names?
How does `remoteBase` relates to `targetRemote` while `targetBase` relates to `sourceRemote`?
It kinda messes with my head even knowing what the code does.

`base` here points to the target branch, so `remoteBase` could be `targetRef` or `targetBase`?
`head` is the source branch, so `targetBase` could be `sourceRef` or `sourceBase`?

bmeneguele (changes requested):

bmeneguele:
@prarit, there still have one place you need to change:
```go
msg, err := mrText(targetBranch, branch, sourceRemote, targetRemote, coverLetterFormat)
```

The `branch` there should be `remoteBranch` instead, since the `branch` points to the local branch and ignores if it has an upstream branch name different from the local. For instance, if I `git push -u <remote> <local-branch-name>:<remote-branch-name>` the changes, in the MR EDIT message the `<local-branch-name>` will be used.

Since `remoteBranch` is being used for calculating the diff, it makes sense to also use its name in the MR EDIT text.

prarit:
> @prarit, there still have one place you need to change:
> 
> ```go
> msg, err := mrText(targetBranch, branch, sourceRemote, targetRemote, coverLetterFormat)
> ```
> 
> The `branch` there should be `remoteBranch` instead, since the `branch` points to the local branch and ignores if it has an upstream branch name different from the local. For instance, if I `git push -u <remote> <local-branch-name>:<remote-branch-name>` the changes, in the MR EDIT message the `<local-branch-name>` will be used.
> 
> Since `remoteBranch` is being used for calculating the diff, it makes sense to also use it in the MR EDIT text.

Good point.  I was running on the assumption that the local source branch name and remote source branch names are the same.  I'll fix that.

bmeneguele (approved):
Yep, LGTM now :))

PR #641, opened 2021-03-20
Fix mr test prepare

prarit:
This was noticed while implementing the --source option in 'mr create'.
    
    The prepare function in the merge request test code has an error that is
    commented out.  This is because the code no longer works correctly and
    always throws an error.  The prepare function attempts to clean up any
    existing test merge requests prior to creating new ones.
    
    Fix and generalize the prepare test function.
    
### DIFF
--- a/cmd/mr_test.go
+++ b/cmd/mr_test.go
@@ -5,6 +5,7 @@ import (
 	"io/ioutil"
 	"os/exec"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"testing"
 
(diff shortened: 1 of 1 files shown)


cmd/mr_test.go
```diff
@@ -5,25 +5,64 @@ import (
...
+	closeMRcmd := exec.Command("sh", "-c", fmt.Sprintf("%s mr close %d\n", labBinaryPath, openMR))
+	closeMRcmd.Dir = cmdDir
+	c, err := closeMRcmd.CombinedOutput()
+	if err != nil {
+		t.Log(string(c))
+		t.Fatal(err)
+	}
+}
+
+func deleteMR(t *testing.T, targetRepo string, cmdDir string, mrID string) {
```
bmeneguele, line 46: I think it would be better if you change `deleteMR` to `closeMR`... I'm implementing the `lab mr delete` command right now and their meaning are quite different. Today we have the `delete` as an alias for `close`, which is really misleading.

bmeneguele (changes requested):

cmd/mr_test.go
prarit: Fixed, and I also renamed the "delete" test functions to "close".

cmd/mr_test.go
```diff
@@ -63,6 +71,13 @@ func Test_mrCmd(t *testing.T) {
 	var mrID string
 	t.Run("prepare", func(t *testing.T) {
 		cleanupMR(t, "lab-testing", repo, "mr title")
+		cleanUpCmd := exec.Command(labBinaryPath, "mr", "close", "4676")
```
bmeneguele, line 74: Was this step added only for closing that MR that was causing the build to fail?
Closing an MR with a hard-coded number doesn't seem correct to be kept.

prarit:
@bmeneguele , OOC when you tested your 'mr delete' modification did you do it against lab-testing !4676?

I am getting an error on !4676, but when I list the MRs there is no 4676.
```

!4678 mr title
!4677 mr title
!4675 Draft: mr by branch
!4674 Test draft merge request
!4673 mr title
```

This could be a GitLab bug.

bmeneguele:
@prarit, I deleted this 4676 after I saw it failing for both your PR and mine (the one enabling `lab mr delete`), but it was after your build failed. It's now failing due to 4702. I think the cleanup code is handling something not in the right way. Let me check your changes again.

cmd/mr_test.go
```diff
@@ -5,25 +5,64 @@ import (
...
+		return
+	}
+
+	if openMR <= 0 {
+		// no open MRs
+		return
+	}
+
+	// close the existing MR
+	closeMRcmd := exec.Command("sh", "-c", fmt.Sprintf("%s mr close %d\n", labBinaryPath, openMR))
```
bmeneguele, line 37: @prarit I think I found the issue: you're not passing the `targetRepo` for this command, so the default one is being used, which isn't the same where the MR is.
If you check the test results the cleanup code tries to delete the MR from projectID `4181224` (zaquestion/test), but the correct should be `5694926` (lab-testing/test)

bmeneguele (changes requested):

cmd/mr_test.go
prarit: Oh geez. That's braindead of me.

bmeneguele: Hahaha.
But why not just use "closeMR" it does exactly the same as this command, only additionally checks for the "close" string, which should cause no harm?

bmeneguele: Ok, I'm going to manually delete the MR 4702 and re-run the test.

bmeneguele:
`cmd/mr_test.go:52:9: cannot use openMR (type int) as type string in argument to closeMR`
:)))

prarit:
> `cmd/mr_test.go:52:9: cannot use openMR (type int) as type string in argument to closeMR`
> :)))

Heh, yeah, just caught that.  I was distracted.  I'll reach out to you on IRC to show you why.

P.

bmeneguele:
Yay! And it finally builds :D.
HOWEVER.. lol... don't hate me, please... could you just add a few more words to the "Fix" patch? :sweat_smile:

prarit:
> Yay! And it finally builds :D.
> HOWEVER.. lol... don't hate me, please... could you just add a few more words to the "Fix" patch? sweat_smile

Yup, I was going to squash that in next.

prarit:
Actually, no.  There's still something wrong with the source test.

bmeneguele (approved):
Thanks @prarit

PR #643, opened 2021-03-20
mr: add delete command

bmeneguele:
This PR adds the ability to delete an MR and also change any occurrence of mr "delete" to mr "close", since `lab mr delete` was being used as an alias to `lab mr close`, which is really misleading considering these are two different actions to GitLab API.


prarit:
LGTM, how about dropping your second patch and let me do that in my cleanup?

bmeneguele:
Yes, sure :)

PR #647, opened 2021-03-24
mr edit: Fix reviewers check

prarit:
The mr edit option --reviewers, incorrectly fails with "aborting: no
changes" because the change to the reviewers is not verified.

Add verification of the reviewers change to the "no changes" check.


prarit:
Quick fix.  I will wait for CI to complete and push.

PR #648, opened 2021-03-24
Update mr_test.go to fix test failure case

prarit:
Fix test failures in cmd/mr_test.go

### DIFF
--- a/cmd/mr_test.go
+++ b/cmd/mr_test.go
@@ -29,7 +29,13 @@ func closeMR(t *testing.T, targetRepo string, cmdDir string, mrID string) {
 }
 
 func cleanupMR(t *testing.T, targetRepo string, cmdDir string, MRtitle string) {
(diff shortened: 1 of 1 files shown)


prarit:
The problem with cmd/mr_test.go:cleanupMR() is that the way it is called is insufficient to clean up a failed test.  The sequence that is leading to failures is, for example,

test run #1:

Test_mrCmd() runs.
Test_mrCmd_MR_description_and_options() runs and fails leaving, for example, a Merge Request with description "Fancy Description".
The test run fails.

test run #2 
Test_mrCmd() runs and fails because a Merge Request with description "Fancy Description" exists.  It is not "found" by the cleanupMR() function because Test_mrCmd()'s call is looking for a description of "mr title".

test run #3 (with my changes above) cleans things up.

test run #4 should succeed.

It's annoying that this takes many tries but it does work.

prarit:
@bmeneguele ready for review (finally, after a lot of debugging).

bmeneguele (approved):

PR #655, opened 2021-03-26
mr_list: Add --reviewer option

prarit:
Add an option, similar to the --author and --assignee search options, that
allows users to search on the GitLab Reviewer field.

Add a --reviewer option to return the list of all merge requests for
which the specified user is listed as a GitLab Reviewer.

### DIFF
--- a/cmd/mr_list.go
+++ b/cmd/mr_list.go
@@ -32,6 +32,8 @@ var (
 	mrAssigneeID   *int
 	order          string
 	sortedBy       string
+	mrReviewer     string
+	mrReviewerID   *int
 )
 
 // listCmd represents the list command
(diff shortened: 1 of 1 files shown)


prarit:
@bmeneguele , next on my list of TODOs is to add testing for the --assignee and --reviewer fields.  I will create MRs (as user prarit) in the lab-testing/test project and add user @lab-testing as a reviewer and assignee.  We should be able to target those MRs for tests.

cmd/mr_list.go
```diff
@@ -109,6 +111,14 @@ func mrList(args []string) ([]*gitlab.MergeRequest, error) {
 		mrMilestone = milestone.Title
 	}
 
+	if mrReviewer != "" {
+		reviewerID, err := lab.UserIDByUserName(mrReviewer)
```
bmeneguele, line 115: It conflicts with the code from MR #650, where I'm removing the `lab.UserIDByUserName` function.
Use the `getUserID` from cmd/utils.go instead.

bmeneguele (changes requested):

bmeneguele (approved):

PR #657, opened 2021-03-28
root: fix git passthrough for unknown flags

bmeneguele:
Flags not being handled by neither `lab` and `git` were printing `git` help command instead of `lab`.
This MR fixes this behavior by checking the `git` result first and returning the execution flow to `lab`.


prarit (approved):
LGTM!

PR #660, opened 2021-04-02
Add internal leveled logger

bmeneguele:
This MR add an internal leveled logger to lab, allowing different levels of log messages, from NONE to DEBUG.
What influenced this additional was the fact the module go-retryablehttp was printing debug messages to the output in case of connection timeouts and/or retries to gitlab server, so some users were complaining about that. For getting rid of it, a custom Leveled Logger should be passed to that through go-gitlab.

With this addition, this MR also adds two root flags: `--debug` and `--quiet`. The first, enable debug messages to be printed, while the second prevents any sort of logging other than the default output from commands.

This PR is still draft, since it depends on the PR https://github.com/xanzy/go-gitlab/pull/1121 to be accepted and merged, since go-gitlab today don't offer a way to pass a custom logger to go-retryablehttp.

### DIFF
--- a/cmd/ci_artifacts.go
+++ b/cmd/ci_artifacts.go
@@ -3,7 +3,6 @@ package cmd
 import (
 	"fmt"
 	"io"
-	"log"
 	"os"
 
 	"github.com/rsteube/carapace"
--- a/cmd/ci_lint.go
+++ b/cmd/ci_lint.go
@@ -3,7 +3,6 @@ package cmd
 import (
 	"fmt"
 	"io/ioutil"
-	"log"
 	"os"
 
 	"github.com/pkg/errors"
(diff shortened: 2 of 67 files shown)


bmeneguele:
The `go-gitlab` change was merged. I´ll test it locally and let you know the results.

bmeneguele:
Manual and automated testing (locally performed) completed just fine.
However, we need to wait until `go-gitlab` releases a new version with the change, thus I can also update `go.mod`.

bmeneguele:
Before merging this MR I'm going to push the tests for this new code.

prarit (approved):

zaquestion (approved):

bmeneguele:
There were two changes since you guys approved: test code added and a fix in the code found during the test development :D

PR #666, opened 2021-04-14
Todo improvements

prarit:
Add todo commands to mark Todo items as Done and all Done, and add a commands to add an MR to a user's Todo list.


prarit:
I'm going to put testing for these features in a separate PR.  I have to think about it some more...

prarit:
My "add" option needs to be expanded to include **both** MRs and Issues.  I'll rework that.

prarit:
> My "add" option needs to be expanded to include **both** MRs and Issues. I'll rework that.

The GitLab API only supports the adding of Merge Requests and issues as TODO items

https://docs.gitlab.com/ce/api/merge_requests.html#create-a-to-do-item
https://docs.gitlab.com/ee/api/issues.html#create-a-to-do-item

but go-gitlab currently only supports the adding of Merge Requests as todos:

https://pkg.go.dev/github.com/xanzy/go-gitlab#MergeRequestsService.CreateTodo

bmeneguele:
I've opened an MR in go-gitlab adding the support for TODOs on issues: https://github.com/xanzy/go-gitlab/pull/1130
Do you prefer to get it accepted there or move forward with MR support only for now?

cmd/todo_mr.go
```diff
@@ -0,0 +1,59 @@
...
+	"os"
+	"strconv"
+
+	"github.com/spf13/cobra"
+	lab "github.com/zaquestion/lab/internal/gitlab"
+)
+
+var todoMRCmd = &cobra.Command{
+	Use:              "mr",
+	Aliases:          []string{"issue"},
```
bmeneguele, line 15: I know it's a "draft" PR, but... hahah..
I don't think that adding one single command to handle both `mr` and `issue` is a good idea :D.
The `lab todo add` approach with flags for defining issue vs mr seems a far better option IMHO.

internal/gitlab/gitlab.go
```diff
@@ -1709,3 +1709,32 @@ func TodoList(opts gitlab.ListTodosOptions, n int) ([]*gitlab.Todo, error) {
...
+
+func MarkAllTodosAsDone() error {
+	_, err := lab.Todos.MarkAllTodosAsDone()
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func CreateTodo(project string, mrNum int) (int, error) {
```
bmeneguele, line 1729: We've been using the not-well-defined pattern of `<object><action>` for function names, e.g. `MRList` or `IssueCreate`, it would be better to keep the pattern here too:

`TodoMarkDone()`
`TodoMarkAllDone()`
`TodoCreate()`.

What do you think?

bmeneguele (changes requested):

cmd/todo_mr.go
prarit: Haha :)  It is definitely a work in progress:

`lab todo add 1234 --mr` : but this is kind of clunky, and that's why I came up with 

`lab todo mr 1234`  which seems like a better idea.

What about

`lab mr add 1234` ? :)  Would that be better?  Or too confusing?

internal/gitlab/gitlab.go
prarit: I like that a lot more and ;) was wondering about the naming.  I'll definitely move to this.

prarit:
> I've opened an MR in go-gitlab adding the support for TODOs on issues: [xanzy/go-gitlab#1130](https://github.com/xanzy/go-gitlab/pull/1130)
> Do you prefer to wait until it gets accepted there, or move this PR forward with MR support only for now?
> Unfortunately, we'll also need to wait until they release a new version, so we can update our go.mod accordingly.

We can continue forward.  Adding Todo support for issues should be trivial.

internal/gitlab/gitlab.go
prarit: Fixed.

cmd/todo_mr.go
prarit: The more I think about this, the more I like the idea of having 

`lab todo add 1234 --mr`

*AND*

`lab mr add 1234`

The second command could be an easy wrapper around the first so there's a minimum of code duplication.

bmeneguele: `lab mr add` is too similar to `lab mr create`, isn't it?
What about `lab todo mr <mrid>` and `lab todo issue <issue-id>` ?

prarit: Yeah, that'd work too.  Let me change the code...

prarit: Let me know what you think of the recent push.

cmd/todo_issue.go
```diff
@@ -0,0 +1,39 @@
...
+import (
+	"log"
+	"strconv"
+
+	"github.com/spf13/cobra"
+)
+
+var todoIssueCmd = &cobra.Command{
+	Use:              "issue",
+	Short:            "Add a Issue to Todo list",
```
bmeneguele, line 12: What do you think about adding a `Hidden: true` in here so this command doesn't show in the usage message?
In this way, we can keep this code and avoid users calling it.

cmd/todo_mr.go
```diff
@@ -0,0 +1,51 @@
...
+		todoAddMergeRequest(rn, num)
+	},
+}
+
+func todoAddMergeRequest(remote string, mrNum int) {
+	todoID, err := lab.TodoMRCreate(remote, mrNum)
+	if err != nil {
+		if err.Error() == "EOF" {
+			log.Fatalf("Todo entry already exists for MR !%d", mrNum)
+		}
```
bmeneguele, line 38: Instead of checking for `EOF`, IMO it would be better to check for the `resp` return code in `lab.TodoMRCreate()` and return a valid error, similar to the approach we used when we created `ErrProjectNotFound` and others.

internal/gitlab/gitlab.go
```diff
@@ -1725,3 +1725,16 @@ func TodoMarkAllDone() error {
...
+func TodoMRCreate(project string, mrNum int) (int, error) {
+	p, err := FindProject(project)
+	if err != nil {
+		return 0, err
+	}
+
+	todo, _, err := lab.MergeRequests.CreateTodo(p.ID, mrNum)
+	if err != nil {
+		return 0, err
+	}
```
bmeneguele, line 1738: Based on my last comment, we should not ignore `resp` return code here: GitLab API states it returns `HTTP 304 Not Modified` for already existent to-do items, in this case we should check `resp.StatusCode == http.StatusNotModified` and return a new error.

bmeneguele (changes requested):

cmd/todo_issue.go
prarit: Good suggestion.  I'll add it although I'm worried we're going to forget about it when we do add support :)

cmd/todo_mr.go
prarit: Good point.  Fixed.

internal/gitlab/gitlab.go
prarit: Fixed as well.

bmeneguele (approved):
LGTM :)

PR #668, opened 2021-04-19
todo: Add --pretty output

prarit:
The 'todo list' output currently only shows the Todo ID, and a link to the
Todo item.  The WebUI outputs additional information such as the state of
the Merge Requests and Issues, how the Todo entry was added to the Todo
list, etc.

Add an option, --pretty, to output the additional information.

### DIFF
--- a/cmd/todo_list.go
+++ b/cmd/todo_list.go
@@ -6,6 +6,7 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/fatih/color"
 	"github.com/spf13/cobra"
 	gitlab "github.com/xanzy/go-gitlab"
 	lab "github.com/zaquestion/lab/internal/gitlab"
@@ -15,6 +16,7 @@ var (
 	todoType   string
 	todoNumRet string
 	targetType string
+	todoPretty bool
 )
 
 var todoListCmd = &cobra.Command{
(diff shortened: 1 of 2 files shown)


prarit:
@bmeneg, the pretty output is a tad slower than the normal output.  This is because the pretty output does an status lookup on each MR and Issue in the Todo list.

prarit:
@bmeneguele after fighting with different formats for this, I decided on keeping the --pretty format as is.  While it doesn't 100% confirm to the Todo list in the WebUI I would argue that because we have to include the TodoID, the web URL, and an explanation *and* I want it in a machine readable format, there's only so many ways to parse the data.

internal/gitlab/gitlab.go
```diff
@@ -383,21 +383,27 @@ func MRCreateNote(project string, mrNum int, opts *gitlab.CreateMergeRequestNote
...
+
+// MRGet retrieves the merge request from GitLab project
+func MRGet(project string, mrNum int) (*gitlab.MergeRequest, error) {
+	p, err := FindProject(project)
 	if err != nil {
 		return nil, err
 	}
 
-	return mr, nil
+	return MRGetFromProjectID(p.ID, mrNum)
```
bmeneguele, line 403: @prarit , I think we can have a single function for getting the project object: the `interface{}` type here is used because it can be either `string` or `int`, it's handled in the go-gitlab module, since the [API](https://docs.gitlab.com/ee/api/projects.html#get-single-project) accepts both (ID as int and escaped project URL path as string).

With that, I think we can reduce the project lookup significantly. But, at the same time, we can leave this "optimization" for a future MR. Your call :D

cmd/todo_list.go
```diff
@@ -33,8 +35,76 @@ var todoListCmd = &cobra.Command{
 		pager := NewPager(cmd.Flags())
 		defer pager.Close()
 
+		red := color.New(color.FgRed).SprintFunc()
```
bmeneguele, line 38: What about using another color (cyan?) for merged MRs instead of red?

bmeneguele (changes requested):

cmd/todo_list.go
prarit: CYAN!?!?!?!?  I HATE CYAN.  Just joking, good idea.  I'll make this change in v2.

internal/gitlab/gitlab.go
prarit: Oh, I didn't realize that about the call.  Let me remove it for now.

prarit: So I fixed this and tested it with podman.  See changes.

cmd/todo_list.go
prarit: Fixed.  Stupid cyan.

internal/gitlab/gitlab.go
bmeneguele: Yep.. Looks good to me. Tested it locally and seems to be working properly as well.

bmeneguele (approved):
This CYAN is AMAZING @prarit .. you made my day just better. Really nice design choice of yours :)))))

PR #674, opened 2021-05-07
Fix assign messages

prarit:
approve/unapprove: Provide better error messages for repeat actions
    
    Users are presented with the following error message if an already
    approved MR is approved again:
    
    2021/05/07 10:34:35 mr_approve.go:55: POST https://gitlab.com/api/v4/projects/24140364/merge_requests/119/approve: 401 {message: 401 Unauthorized}
    
    Similarily, when an unapproved MR is unapproved again:
    
    2021/05/07 10:35:01 mr_unapprove.go:55: POST https://gitlab.com/api/v4/projects/24140364/merge_requests/119/unapprove: 40
    
    These messages are confusing and should state that the MR was already
    approved or unapproved.
    
    Add better error messages for repeat approve and unapprove actions.
    

prarit:
@bmeneguele,  I'm breaking the gitlab.go error codes :) intentionally to provide a better error status.  Also, I can be very easily convinced to change the successful os.Exit(0)'s to os.Exit(1)'s.

internal/gitlab/gitlab.go
```diff
@@ -538,6 +545,10 @@ func MRUnapprove(pid interface{}, id int) error {
 	if resp != nil && resp.StatusCode == http.StatusForbidden {
 		return ErrStatusForbidden
 	}
+	if resp != nil && resp.StatusCode == http.StatusNotFound {
+		// returns 404 if the MR has already been approved
```
bmeneguele, line 549: s/approved/unapproved maybe? :)

Just curious, the API returns 401 when it was already approved and 404 when already unapproved? two different error codes for basically the same operation? Don't get me wrong, I'm not doubting or anything like that, I'm just surprised how odd it is.... even more using 404 (not found??). lol.

bmeneguele (changes requested):

internal/gitlab/gitlab.go
prarit: Unapprove makes sense. I'll change that comment.

The 401/404 makes sense.  Think of this it this way: in order to have an unapprove available the MR must be in the approve state.  If the MR is in the unapprove state then the unapprove does not exist -- hence 404 (Not Found).  If an MR is approved, then someone attempts to approve it again, then they should be blocked from approving a second time -- hence 401 (Unauthorized).

prarit: Do you think I should return "fail" (ie "1") in these cases instead of success?  I mean -- the command to approve/unapprove didn't actually do anything and that should be a failure, no?

PR #675, opened 2021-05-10
gitlab: fix the number of items per page

bmeneguele:
Instead of requesting multiple pages with the default number of items (20),
or requesting more than the max allowed by the API, request as much as
possible items per page, avoiding multiple calls in "list" subcommands.


prarit (approved):

bmeneguele:
OK. after _days_ trying to understand why the testing code was "timing out" I discovered an infinite loop in the code.
The pagination loop was being tested against a local value instead of using the response structure already with a refreshed value, with that, we needed to keep refreshing our own to keep things in sync... which doesn't make much sense. So I added another patch to make use of the `Response` object.

PR #679, opened 2021-05-20
todo list: Remove MR and Issue lookups

prarit:
While looking at some other code I discovered that type Todo contains
a type TodoTarget which contains the information that I was pulling
from the MR and Issue lookups.  These lookups can be dropped and the
code is simplified, there are fewer API calls, and 'lab todo list
--pretty' is faster.

Remove the MR and Issue lookups with the --pretty option.


bmeneguele (approved):
Ohh nice! :)

PR #682, opened 2021-05-20
Fix build for BSD OSes family

### LINKED ISSUE #680: v0.21.0 fails to build on FreeBSD
I'm seeing a build failure on FreeBSD:

```
github.com/zaquestion/lab/cmd
# github.com/zaquestion/lab/cmd
cmd/util_unix.go:33:9: undefined: syscall.Dup3
```
Yes, the dup3 syscall exists, but I think you can't access it from go via `syscall`. I think you have to use `x/sys/unix` Syscall()

bmeneguele:
This merge request fixes #680 by re-organizing the cmd/util_* files for specific GOOS.
Darwin and *BSD (mostly FreeBSD) share some ground when it comes to kernel API, with that, a good starting point is to add both families in the same file, suffixed by _bsd.
I know it doesn't hold true for every syscall, but for what we need in `lab` it's good enough.

### DIFF
--- a/cmd/util_bsd.go
+++ b/cmd/util_bsd.go
@@ -1,6 +1,6 @@
 // This file contains Darwin (MacOS) and *BSD specific calls.
 
-// +build darwin
+// +build freebsd openbsd dragonfly darwin
 
 package cmd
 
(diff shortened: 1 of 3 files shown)


bmeneguele:
Test done by the issue reporter.

PR #686, opened 2021-06-10
Make use of heredoc.Doc() module in multiline texts

bmeneguele:
We have multiple places where multiline texts are used within the actual code (help messages, text templates, etc..).
Using the usual back-quote marks allow that, but at the same time it considers all the leading whitespaces in each line, making it look really ugly.

This MR adds the use of `heredoc.Doc()` module/function to solve it, which considers valid leading whitespace only what is beyond the first line indentation level. Tests multiline texts were not changed to keep checking the `heredoc.Doc()`  behavior.

Also, this MR updates some help messages for some commands and doesn't contain any functional code changes.

### DIFF
--- a/cmd/ci.go
+++ b/cmd/ci.go
@@ -11,7 +11,6 @@ var followBridge bool
 var ciCmd = &cobra.Command{
 	Use:   "ci",
 	Short: "Work with GitLab CI pipelines and jobs",
-	Long:  ``,
 }
 
 func init() {
--- a/cmd/ci_artifacts.go
+++ b/cmd/ci_artifacts.go
@@ -5,6 +5,7 @@ import (
 	"io"
 	"os"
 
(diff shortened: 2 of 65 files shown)


bmeneguele:
One specific build failure was caused by the artifacts expiration date of a single CI job.
I retried it (generated a new version of the same job) and set the expiration date to "never:
```
curl -X POST -H "PRIVATE-TOKEN: <lab-testing key>" 'https://gitlab.com/api/v4/projects/4181224/jobs/1339905570/artifacts/keep'
```
Hopefully we won't face the same issue again, unless someone rerun the job without setting the expiration date.

bmeneguele:
Since this MR contains only doc changes I'm going to merge it.
If something (typos) are noted later we can easily create new MRs.

PR #691, opened 2021-06-15
mr show: Fix --patch and --comments option output

prarit:
Users have been setting "comments = true" for the mr show command.  This
causes weird output with the --patch output in which the git log is output,
then the comments from the MR is output.

Make a note that --patch and --comments do not work together and stop
the output after the git log is output.

### DIFF
--- a/cmd/mr_show.go
+++ b/cmd/mr_show.go
@@ -62,6 +62,7 @@ var mrShowCmd = &cobra.Command{
 				log.Fatal(err)
 			}
 			git.Show(remote+"/"+mr.TargetBranch, mr.SHA, mrShowPatchReverse)
+			return
 		} else {
 			pager := newPager(cmd.Flags())
 			defer pager.Close()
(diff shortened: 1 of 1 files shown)


prarit:
So I'm one of these users who have set "comments = true" :) and it's been a pain for a whlie now.  Finally got around to fixing this :)

cmd/mr_show.go
```diff
@@ -239,9 +240,9 @@ func printMR(mr *gitlab.MergeRequest, project string, renderMarkdown bool) {
 
 func init() {
 	mrShowCmd.Flags().BoolP("no-markdown", "M", false, "don't use markdown renderer to print the issue description")
-	mrShowCmd.Flags().BoolP("comments", "c", false, "show comments for the merge request")
+	mrShowCmd.Flags().BoolP("comments", "c", false, "show comments for the merge request (does not work with --patch)")
 	mrShowCmd.Flags().StringP("since", "s", "", "show comments since specified date (format: 2020-08-21 14:57:46.808 +0000 UTC)")
-	mrShowCmd.Flags().BoolVarP(&mrShowPatch, "patch", "p", false, "show MR patches")
+	mrShowCmd.Flags().BoolVarP(&mrShowPatch, "patch", "p", false, "show MR patches (does not work with --comments)")
```
bmeneguele, line 245: What if: instead of warning it in the help message, you actually block it in code level, returning error if both are set.
Like we do for `lab mr edit|create` with `-F` and `-m` being used together.

prarit: Good idea :) but that doesn't work because if I set comments = true in the .config, then "show -p" would fail 100% of the time.

bmeneguele: Fair point.
Would that make sense to print an WARNING at least?
In the help message it seems pretty easy to miss, that's why I'm asking for some runtime feedback from lab to the user :)

prarit: I'd still end up seeing the warning everytime I ran it :(

bmeneguele: Well.. ok then :)
I'm fine with it.

PR #694, opened 2021-06-22
cmd/mr_show: golint: if block ends with a return statement

bmeneguele:
Remove the else clause considering the if ends with return.


zampierilucas:
LGTM, ack

PR #697, opened 2021-06-24
Fix commit log and diffstat when creating MR

bmeneguele:
When the user creates an MR he's prompted with some basic information about what's going to be the MR content:
* commit list
* number of commits
* commit diff statistics

All these information were being created using a "union" approach of `git`: when the information being considered is the union of two different sets (branches) instead of the difference from one set to another (i.e. what's different from branch `<source>` to `<target>`).

The consequence of the union approach appears when the user wants to create an MR using a source branch that is behind of the target branch wrt commits: if the target is 4 commits ahead of the source branch, and the user doesn't rebase before opening the MR, the commits information (all three of them) will present the source branch's commits + 4 from target branch, which is, obviously wrong when compared to what the UI presents during MR creation.

All the information and related commands was took from Git's manpage.

EDIT:

The commit description was also being retrieved from HEAD instead of the specific source remote and branch. Additional commits were added to handle that as well.

### DIFF
--- a/cmd/mr_create.go
+++ b/cmd/mr_create.go
@@ -292,7 +292,7 @@ func mrText(sourceRemote, sourceBranch, targetRemote, targetBranch string, cover
 	numCommits := git.NumberCommits(target, source)
 	if numCommits == 1 {
 		var err error
-		commitMsg, err = git.LastCommitMessage()
+		commitMsg, err = git.LastCommitMessage(source)
 		if err != nil {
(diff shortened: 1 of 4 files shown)


zampierilucas (approved):

bmeneguele:
Something strange is going on with the testing.
Will investigate.

zampierilucas (approved):

PR #702, opened 2021-06-28
cmd/util: fix handle of unknown number of arguments

bmeneguele:
Instead of returning an error in case of an unknown number of args, let
it return the nil values and let the caller handle it as it please. For
"mr_note", "issue_note" and "mr_edit" the error was interrumpting the
cmd line argument parser.

Fixes: ecd57aaf4afa ("cmd/util: avoid an empty string to be parsed")


fmuellner:
Ah, I was about to report this :-)

LGTM

PR #703, opened 2021-06-28
Move noteRunFn to a common place and fix os.Args usage

bmeneguele:
This MR clean 2 things up:

1) move noteRunFn to cmd/note_common.go file since it's used by both mr_note and issue_note;
2) instead of using `os.Args`, use `cmd.Parent().Name()` and `cmd.CalledAs()` to avoid surprises with `os.Args` being mangled.

### DIFF
--- a/cmd/issue_note.go
+++ b/cmd/issue_note.go
@@ -1,9 +1,6 @@
 package cmd
 
 import (
-	"fmt"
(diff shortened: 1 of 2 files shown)


fmuellner (approved):
LGTM

PR #704, opened 2021-06-29
chore: (ci status) #240 bump up default wait interval

zaquestion:
Per discussion in #240, lets just bump up the interval, totally open to putting it behind a flag, but in the context of waiting for a an entire pipeline to finish, 1s is clearly a bit too aggressive.


zaquestion:
Yeah, exactly

PR #705, opened 2021-06-30
mr show: Output commit ID hash on commit comments

prarit:
Output the commid ID hash on commit comments.

Note: The behavior of note.Position.BaseSHA has changed to point to the
commit just before the commented on commit.


bmeneguele:
@prarit I'm not sure if I understand the issue here. could you show an example just to help me understand/reproduce?

bmeneguele:
Ok, I see what's the point of it now.

bmeneguele (approved):

PR #707, opened 2021-07-01
Removed --with-comment dependency; Fixed approve bug

zampierilucas:
Fixed a bug where a comment would be added to an MR on approve or
unapprove, even if the MR was already reviewed by the user.

Changed -m and -f behavior so they don't depend on --with-comment to be
used.

### DIFF
--- a/cmd/mr_approve.go
+++ b/cmd/mr_approve.go
@@ -30,23 +30,15 @@ var mrApproveCmd = &cobra.Command{
 		if err != nil {
 			log.Fatal(err)
 		}
-		if comment {
-			msgs, err := cmd.Flags().GetStringArray("message")
-			if err != nil {
-				log.Fatal(err)
-			}
 
-			filename, err := cmd.Flags().GetString("file")
-			if err != nil {
-				log.Fatal(err)
-			}
+		msgs, err := cmd.Flags().GetStringArray("message")
+		if err != nil {
+			log.Fatal(err)
+		}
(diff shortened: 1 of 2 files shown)


cmd/mr_approve.go
```diff
@@ -59,15 +51,25 @@ var mrApproveCmd = &cobra.Command{
 				os.Exit(1)
 			}
 		}
+
+		if comment || len(msgs) > 0 || filename != "" {
```
bmeneguele, line 55: Maybe it's worth adding a `strings.TrimSpace(filename)`  here?

bmeneguele: ah yea.. the same would be valid in the mr_unapprove.go

zampierilucas: You're right, a new commit with the added trim has pushed :)

bmeneguele:
you know what.. that's my bad. After some IRC discussion I agree with your first version: we should not try to correct user's bad behavior. we just need to make sure an empty filename doesn't break lab (panic) somewhere, but if the user passed an invalid filename the editor should be the one reporting "invalid filename". lab should do what the users asked for.. if that's opening a weird filename, lets try it.

With that said, yeah, feel free to remove the last commit that you added the `strings.TrimSpace()`.
Sorry for that.

bmeneguele (approved):

prarit (approved):

PR #711, opened 2021-07-06
Fix thumb Down

zampierilucas:
`Thumb Down` cannot determine MR id, changed the function to the same used on `Thumb Up`


bmeneguele (approved):
Há! Good catch.
Thanks @zampierilucas

PR #716, opened 2021-07-07
cmd/snippet_browse: Fix Snippets Browse cfg host

zampierilucas:
Snippets could not get hostURL from cfg, mimicked the code from
mr_browse.


bmeneguele (approved):
Ok, this predates the config overhaul we did some time ago.
Good catch.

PR #719, opened 2021-07-27
fix: (ci view) dont run ci job if `esc` is hit

zaquestion:


bmeneguele (approved):
Thanks for taking care of it @zaquestion :) I've been on vacation for the past 2 weeks and still have 1 to go, even though I was planning to work for a bit on the issues, I couldn't :D.

PR #722, opened 2021-08-05
Add commit comment functionality

prarit:
The GitLab WebUI allows users to comment directly on commits.  These comments
    appear both in the comment view and the MR view.  This functionality can be
    implemented via the lab.Commits.PostCommitComment() function, however, the
    function only supports comments on "new" lines of code.  After discussions
    with GitLab they have suggested using lab.Discussions.CreateCommitDiscussion()
    as a workaround [1].  This solution has a minor issue in that it will not work
    with commits that have multiple parents
    
    Implement the ability for users to comment directly on commits.
    
    [1] https://gitlab.com/gitlab-org/gitlab/-/issues/335337#note_643291819

### DIFF
--- a/cmd/issue_note.go
+++ b/cmd/issue_note.go
@@ -28,6 +28,8 @@ func init() {
 	issueNoteCmd.Flags().Bool("quote", false, "quote note in reply")
 	issueNoteCmd.Flags().Bool("resolve", false, "[unused in issue note command]")
 	issueNoteCmd.Flags().MarkHidden("resolve")
+	issueNoteCmd.Flags().StringP("commit", "", "", "[unused in issue note command]")
+	issueNoteCmd.Flags().MarkHidden("commit")
 
 	issueCmd.AddCommand(issueNoteCmd)
 	carapace.Gen(issueNoteCmd).PositionalCompletion(
--- a/cmd/mr_discussion.go
+++ b/cmd/mr_discussion.go
@@ -25,7 +25,8 @@ var mrCreateDiscussionCmd = &cobra.Command{
 	Example: heredoc.Doc(`
 		lab mr discussion origin
 		lab mr discussion my_remote -m "discussion comment"
-		lab mr discussion upstream -F test_file.txt`),
+		lab mr discussion upstream -F test_file.txt
+		lab mr discussion --commit abcdef123456`),
 	PersistentPreRun: labPersistentPreRun,
 	Run: func(cmd *cobra.Command, args []string) {
 		rn, mrNum, err := parseArgsWithGitBranchMR(args)
@@ -43,6 +44,11 @@ var mrCreateDiscussionCmd = &cobra.Command{
 			log.Fatal(err)
 		}
 
+		commit, err := cmd.Flags().GetString("commit")
+		if err != nil {
+			log.Fatal(err)
+		}
+
 		body := ""
 		if filename != "" {
 			content, err := ioutil.ReadFile(filename)
@@ -50,12 +56,21 @@ var mrCreateDiscussionCmd = &cobra.Command{
 				log.Fatal(err)
 			}
 			body = string(content)
-		} else {
+		} else if commit == "" {
 			body, err = mrDiscussionMsg(msgs)
 			if err != nil {
 				_, f, l, _ := runtime.Caller(0)
 				log.Fatal(f+":"+strconv.Itoa(l)+" ", err)
 			}
+		} else {
+			body = getCommitBody(rn, commit)
+			body, err = noteMsg(nil, true, body)
+			if err != nil {
+				_, f, l, _ := runtime.Caller(0)
+				log.Fatal(f+":"+strconv.Itoa(l)+" ", err)
+			}
+			createCommitComments(rn, int(mrNum), commit, body, true)
+			return
 		}
 
 		if body == "" {
@@ -117,6 +132,7 @@ func mrDiscussionText() (string, error) {
 func init() {
 	mrCreateDiscussionCmd.Flags().StringSliceP("message", "m", []string{}, "use the given <msg>; multiple -m are concatenated as separate paragraphs")
 	mrCreateDiscussionCmd.Flags().StringP("file", "F", "", "use the given file as the message")
+	mrCreateDiscussionCmd.Flags().StringP("commit", "c", "", "start a thread on a commit")
 
 	mrCmd.AddCommand(mrCreateDiscussionCmd)
 	carapace.Gen(mrCreateDiscussionCmd).PositionalCompletion(
--- a/cmd/mr_note.go
+++ b/cmd/mr_note.go
@@ -17,7 +17,8 @@ var mrNoteCmd = &cobra.Command{
 		lab mr note a_remote -F test_file --force-linebreak
 		lab mr note upstream -m "A helpfull comment"
 		lab mr note upstream:613278106 --quote
-		lab mr note upstream:613278107 --resolve`),
+		lab mr note upstream:613278107 --resolve
+		lab mr note upstream:613278108 --commit abcdef123456`),
 	PersistentPreRun: labPersistentPreRun,
 	Run:              noteRunFn,
 }
@@ -28,6 +29,7 @@ func init() {
 	mrNoteCmd.Flags().Bool("force-linebreak", false, "append 2 spaces to the end of each line to force markdown linebreaks")
 	mrNoteCmd.Flags().Bool("quote", false, "quote note in reply")
 	mrNoteCmd.Flags().Bool("resolve", false, "mark thread resolved")
+	mrNoteCmd.Flags().StringP("commit", "c", "", "start a thread on a commit")
 
 	mrCmd.AddCommand(mrNoteCmd)
 	carapace.Gen(mrNoteCmd).PositionalCompletion(
--- a/cmd/note_common.go
+++ b/cmd/note_common.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	"bufio"
 	"bytes"
 	"fmt"
 	"io/ioutil"
@@ -66,6 +67,11 @@ func noteRunFn(cmd *cobra.Command, args []string) {
 		log.Fatal(err)
 	}
 
+	commit, err := cmd.Flags().GetString("commit")
+	if err != nil {
+		log.Fatal(err)
+	}
+
 	if reply != 0 {
 		resolve, err := cmd.Flags().GetBool("resolve")
 		if err != nil {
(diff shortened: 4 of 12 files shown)


prarit:
@bmeneguele @zampierilucas  Ready for review.  Had some test failures to fix, and an interesting code panic that testing caught.

prarit:
I just figured out how to create a blocking thread on a commit comment:

```
curl --request POST --header "PRIVATE-TOKEN: <your secret token>" \
--form "body=Friday Aug 6" \
--form "commit_id=e55845bedfe55a439cea4eab508998f85bef7b5d" \
--form "position[old_path]=Makefile" \
--form "position[new_path]=Makefile" \
--form "position[old_line]=461" \
--form "position[base_sha]=f277a231238e6d1093652408f61dd5430b2cfaf9" \
--form "position[start_sha]=f277a231238e6d1093652408f61dd5430b2cfaf9" \
--form "position[head_sha]=e55845bedfe55a439cea4eab508998f85bef7b5d" \
--form "position[position_type]=text" \
-- https://gitlab.com/api/v4/projects/21681515/merge_requests/2/discussions
```

I'm going to add this functionality as well.

bmeneguele:
> I have chosen to keep the "|" instead of ">" suggested by @bmeneguele . As I discovered, the ">" can be used when you are quoting text in Markdown.

:disappointed: .... that's fine though.

Will review it tomorrow. But thanks for getting it done Prarit :)

cmd/note_common.go
```diff
@@ -85,10 +91,122 @@ func noteRunFn(cmd *cobra.Command, args []string) {
...
+			for _, f := range fs {
+				if f == "" || f == "|" {
+					continue
+				}
+				val, err := strconv.Atoi(f)
+				if err != nil {
+					// NaN
+					if f == "@@" {
+						continue
+					}
```
zampierilucas, line 186: maybe f == "@@"  could be also checked above?
```suggestion
				if f == "" || f == "|" || f == "@@" {
					continue
				}
				val, err := strconv.Atoi(f)
				if err != nil {
					// NaN
```

prarit: Good suggestion.  I've incorporated into v2.

prarit:
@bmeneguele any chance of a review?

bmeneguele:
Oh yes. Will review it tomorrow. :)

cmd/note_common.go
```diff
@@ -111,7 +226,12 @@ func createNote(rn string, isMR bool, idNum int, msgs []string, filename string,
 				"merged": "MERGED",
 			}[mr.State]
 
-			body = fmt.Sprintf("\n# This comment is being applied to %s Merge Request %d.", state, idNum)
+			if commit != "" {
+				body = getCommitBody(rn, commit)
+				body += fmt.Sprintf("\n# This comment is being applied to %s Merge Request %d commit %s.\n# Do not delete patch tracking lines that begin with '|'.", state, idNum, commit[:12])
```
bmeneguele, line 231: Forcing the user to pass a SHA of 12 chars seems a bit "overhead", considering the GitLab interface, by default, display all commits using 8 chars instead:

image
image

What if we make it, at least, 8 chars?

cmd/note_common.go
```diff
@@ -111,7 +226,12 @@ func createNote(rn string, isMR bool, idNum int, msgs []string, filename string,
 				"merged": "MERGED",
 			}[mr.State]
 
-			body = fmt.Sprintf("\n# This comment is being applied to %s Merge Request %d.", state, idNum)
+			if commit != "" {
```
bmeneguele, line 229: We need to check `len(commit)` here to for the minimum we are requiring, otherwise it'll panic in case the user pass something smaller than we expect. Or the check can be even earlier, when the `commit` flag is retrieved in `noteRunFn()`.

bmeneguele (changes requested):

cmd/issue_note.go
```diff
@@ -28,6 +28,8 @@ func init() {
 	issueNoteCmd.Flags().Bool("quote", false, "quote note in reply")
 	issueNoteCmd.Flags().Bool("resolve", false, "[unused in issue note command]")
 	issueNoteCmd.Flags().MarkHidden("resolve")
+	issueNoteCmd.Flags().StringP("commit", "", "", "[unused in issue note command]")
+	issueNoteCmd.Flags().MarkHidden("commit")
```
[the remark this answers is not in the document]
prarit: The problem is that noteRunFn() will fail to compile because commit variable is undefined.  So I must define the function in issue_note.go.

cmd/note_common.go
prarit: I think 8 chars is not enough.  GitLab in the webUI uses 8 for "pretty" reasons.  The problem is that I've used less than 10 and had git errors indicating that the sha is ambiguous.
For example: 
```

error: short object ID abcd is ambiguous
hint: The candidates are:
hint:   abcd083a1a65 commit 2008-11-25 - sched: convert sched.c from for_each_cpu_mask to for_each_cpu.
hint:   abcd08a6f564 commit 2007-05-09 - SLUB: use check_valid_pointer in kmem_ptr_validate
hint:   abcd3d6fc640 commit 2018-05-04 - net: nixge: Fix error path for obtaining mac address
hint:   abcd7789118c commit 2021-01-14 - [netdrv] ib
```

cmd/note_common.go
prarit: Will do.  I will check it noteRunFn() to be consistent.

prarit: Fixed in update.  Now errors out with:
```
[prarit@prarit rhel-8]$ ~/git-kernel/github/lab/lab mr note 649 --commit 8b37ff4a
2021/08/19 15:52:46 ERROR: note_common.go:76: Error: commit string length must be greater than 12 characters.
```

cmd/issue_note.go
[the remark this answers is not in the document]
zampierilucas: @prarit I've faced the same issue when I added L38, would love to find an alternative to that tho.
Maybe we can create a parent class definition for `issue` and another for `mr`?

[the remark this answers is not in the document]
bmeneguele: Ahh.. `noteRunFn` is being used in both `mrNote` and `issueNote`. I see!
Instead of creating a parent definition for both, we could maybe create a more granular function for the common tasks there? So the flag variables could, at least, be populated by each command `Run:` separated.
But ok, lets not bother with it now. With the explanation I think it's fine.

cmd/note_common.go
bmeneguele: hmmm.. less then 10 chars is already conflicting? That's bad.
But yeah, considering the kernel code.. I don't doubt it :)

What if..... you'll hate me for this: let 8 as default, if git returns "ambiguity" we warn the user to increase the number to at least 12 chars?

I say that because most people using `lab` won't be handling kernel code, but other project far simpler and smaller, so 8 should be a good amount for 90% of the projects out there.

If you prefer, we can let this fix for a follow-up submission, but I think it's important.

cmd/issue_note.go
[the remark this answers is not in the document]
prarit: @zampierilucas, not a bad idea.  I've been thinking about alternatives as well.  I'll look into it -- no promises though :)

cmd/note_common.go
prarit: Good idea!  I'm too kernel-centric in my thinking :)   I've modified the code to error with, for example,

```
[prarit@prarit Test-Lab]$  ~/git-kernel/github/lab/lab mr note 2 --commit abcde
2021/08/19 19:28:41 ERROR: gitlab.go:1803: Cannot find diff for commit abcde.  Verify the commit ID or add more characters to the commit ID.
```

cmd/mr_note.go
```diff
@@ -28,6 +29,7 @@ func init() {
 	mrNoteCmd.Flags().Bool("force-linebreak", false, "append 2 spaces to the end of each line to force markdown linebreaks")
 	mrNoteCmd.Flags().Bool("quote", false, "quote note in reply")
 	mrNoteCmd.Flags().Bool("resolve", false, "mark thread resolved")
+	mrNoteCmd.Flags().StringP("commit", "", "", "start a thread on a commit")
```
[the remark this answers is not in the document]
prarit: I've added a '-c'.  I completely forgot to do so when I wrote the code.

cmd/note_common.go
bmeneguele: I would add "Cannot find diff for commit abcde _or it is ambiguous_".
But that's close enough :)

prarit: Yeah I wanted something a bit more "human"  to read.  Blame my work on the other tooling lately :)

bmeneguele:
@prarit you mentioned you did solved the question, but they weren't pushed yet, right?
Also, don't forget to drop the 'draft' state of the PR :)

prarit:
> @prarit you mentioned you did solved the question, but they weren't pushed yet, right?
> Also, don't forget to drop the 'draft' state of the PR :)

I pushed last night?  At least it shows up in my tree as being pushed:

```
[prarit@prarit ~]$ cd git-kernel/github/lab/
[prarit@prarit lab]$ git where
commit_work2
[prarit@prarit lab]$ git push -f prarit commit_work2
Everything up-to-date
[prarit@prarit lab]$ 
```

and AFAICT the new code is present in the MR:
```

--- a/internal/gitlab/gitlab.go
+++ b/internal/gitlab/gitlab.go
@@ -1799,7 +1799,9 @@ func GetCommitDiff(project string, sha string) ([]*gitlab.Diff, error) {
        for {
                ds, resp, err := lab.Commits.GetCommitDiff(p.ID, sha, opt)
                if err != nil {
-                       fmt.Println(err)
+                       if resp.StatusCode == 404 {
+                               log.Fatalf("Cannot find diff for commit %s.  Verify the commit ID or add more characters to the commit ID.", sha)
+                       }
                        return nil, err
                }
```

cmd/mr_discussion.go
```diff
@@ -117,6 +132,7 @@ func mrDiscussionText() (string, error) {
 func init() {
 	mrCreateDiscussionCmd.Flags().StringSliceP("message", "m", []string{}, "use the given <msg>; multiple -m are concatenated as separate paragraphs")
 	mrCreateDiscussionCmd.Flags().StringP("file", "F", "", "use the given file as the message")
+	mrCreateDiscussionCmd.Flags().StringP("commit", "", "", "start a thread on a commit")
```
bmeneguele, line 135: It's missing the same "-c" option in the `StringP()`.

bmeneguele (changes requested):

bmeneguele:
> > @prarit you mentioned you did solved the question, but they weren't pushed yet, right?
> > Also, don't forget to drop the 'draft' state of the PR :)
> 
> I pushed last night? At least it shows up in my tree as being pushed:
> 
> ```
> [prarit@prarit ~]$ cd git-kernel/github/lab/
> [prarit@prarit lab]$ git where
> commit_work2
> [prarit@prarit lab]$ git push -f prarit commit_work2
> Everything up-to-date
> [prarit@prarit lab]$ 
> ```
> 
> and AFAICT the new code is present in the MR:
> 
> ```
> 
> --- a/internal/gitlab/gitlab.go
> +++ b/internal/gitlab/gitlab.go
> @@ -1799,7 +1799,9 @@ func GetCommitDiff(project string, sha string) ([]*gitlab.Diff, error) {
>         for {
>                 ds, resp, err := lab.Commits.GetCommitDiff(p.ID, sha, opt)
>                 if err != nil {
> -                       fmt.Println(err)
> +                       if resp.StatusCode == 404 {
> +                               log.Fatalf("Cannot find diff for commit %s.  Verify the commit ID or add more characters to the commit ID.", sha)
> +                       }
>                         return nil, err
>                 }
> ```

Yeah, sorry. Had to refresh the browser tab :facepalm:

cmd/mr_discussion.go
prarit: Oops.  Fixed there too!

bmeneguele (approved):
It's looking good to me know :)

PR #723, opened 2021-08-06
Fix gitlab profile link URL

zampierilucas:
Gitlab profile URL must have changed sometime in the past.
Correcting to the new one.


bmeneguele (approved):
Good catch.

PR #733, opened 2021-08-20
Issue732

prarit:
When you start a comment with 'mr note' lab adds, for example:

```
# This comment is being applied to OPEN Merge Request 1166 commit ee6d5ceeadc8f6d8e71a565bb50fcff5e98992fa.
# Do not delete patch tracking lines that begin with '|'.
# Write a message for this note. Commented lines are discarded.
```

And when you start a discussion with 'mr discussion' lab adds, for example:

```
# Write a message for this note. Commented lines are discarded.
```

Make the messaging consistent and unify the code.

Closes 732

### DIFF
--- a/cmd/mr_discussion.go
+++ b/cmd/mr_discussion.go
@@ -49,6 +49,17 @@ var mrCreateDiscussionCmd = &cobra.Command{
 			log.Fatal(err)
 		}
 
+		mr, err := lab.MRGet(rn, int(mrNum))
+		if err != nil {
+			log.Fatal(err)
+		}
+
+		state := map[string]string{
+			"opened": "OPEN",
+			"closed": "CLOSED",
+			"merged": "MERGED",
+		}[mr.State]
+
 		body := ""
 		if filename != "" {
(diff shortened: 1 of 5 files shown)


cmd/mr_discussion.go
```diff
@@ -49,6 +47,17 @@ var mrCreateDiscussionCmd = &cobra.Command{
...
+		if err != nil {
+			log.Fatal(err)
+		}
+
+		state := map[string]string{
+			"opened": "OPEN",
+			"closed": "CLOSED",
+			"merged": "MERGED",
+		}[mr.State]
+
```
bmeneguele, line 60: Why not calling `noteGetState()` here with `isMR = true` instead of replicating the code?

bmeneguele (changes requested):

cmd/mr_discussion.go
prarit: #oops.  I thought I had done that.  Fixed in new version.

bmeneguele (approved):

PR #734, opened 2021-08-24
Remove redundant code and fix flag option

bmeneguele:
This MR has three different and minor fixes:

1. Remove redundant code in `internal/gitlab/gitlab.go::FindProject()`
2. The `--project` value was being used as the value for `token` variable instead of `--token` value.

Check the individual commits for more information.

### DIFF
--- a/cmd/ci_run.go
+++ b/cmd/ci_run.go
@@ -69,7 +69,7 @@ var ciTriggerCmd = &cobra.Command{
 		if err != nil {
 			log.Fatal(err)
 		}
-		token, err := cmd.Flags().GetString("project")
+		token, err := cmd.Flags().GetString("token")
 		if err != nil {
 			log.Fatal(err)
 		}
--- a/internal/gitlab/gitlab.go
+++ b/internal/gitlab/gitlab.go
@@ -197,10 +197,7 @@ func FindProject(project string) (*gitlab.Project, error) {
 		search = user + "/" + project
 	}
 
-	target, resp, err := lab.Projects.GetProject(search, nil)
-	if resp != nil && resp.StatusCode == http.StatusNotFound {
-		return nil, ErrProjectNotFound
-	}
+	target, err := GetProject(search)
 	if err != nil {
 		return nil, err
 	}


cmd/ci_run.go
```diff
@@ -56,20 +52,16 @@ var ciTriggerCmd = &cobra.Command{
...
-		lab ci trigger -p engineering/integration_tests -v foo=bar master`),
+		lab ci trigger -v foo=bar feature_branch`),
 	PersistentPreRun: labPersistentPreRun,
 	Run: func(cmd *cobra.Command, args []string) {
 		pid, branch, err := getCIRunOptions(cmd, args)
 		if err != nil {
 			log.Fatal(err)
 		}
-		token, err := cmd.Flags().GetString("project")
+		token, err := cmd.Flags().GetString("token")
```
zaquestion, line 64: So this line definitely looks suspicious, and I'm particularly nervous since I know I tested triggering ci pipelines. It was a manual test however, there isn't automation for that. This does look like a reasonable fix tho.

cmd/ci_run.go
```diff
@@ -143,12 +135,14 @@ func parseCIVariables(vars []string) (map[string]string, error) {
 
 func init() {
 	ciCreateCmd.Flags().StringP("project", "p", "", "project to create pipeline on")
+	ciCreateCmd.Flags().MarkDeprecated("project", "project is inferred from branch only")
```
zaquestion, line 138: So despite the strangeness with `"project"` being pulled in by `"token"`, the `--project` flag is still relevant in `getCIRunOptions` in this file, and I think it actually behaves correctly there, we have tests for that one. It's also an important feature that folks can trigger ci pipelines in other projects, as lab is used in combination with the `lab ci status --wait` command to orchestrate downstream pipelines (unsure how many people do this, but that's the work flow we built for)

cmd/ci_run.go
zaquestion: Here's a link to the original PR, I think this indeed was just a bug/copypasta error https://github.com/zaquestion/lab/pull/201/files

cmd/ci_run.go
zaquestion: The workflow itself and ideas there stem from https://github.com/zaquestion/lab/issues/240

zaquestion (dismissed):
I think we don't actually need to deprecate the `--project` flag see comments and `getCIRunOptions`

zaquestion:
Ah I'm realizing that marking for deprecation doesn't disable the functionality, tho I still have some concerns about removing it down the line. Despite the inconsistency in the tool, very fair assessment there, the `--project` flag serves a real purpose and powers a real workflow, which is consistent with the goals of the project ( as in it's not there just for api completeness, it's there because it's useful for doing a particular thing). At the least I'd suggest we open an issue to discuss how to support this in a way more consistent with the repo. I would also check what the `lab ci status` command does as I believe I added a way to specify project for that as part of #240

zaquestion (commented):
downgrading my request for changes to a request for comment

cmd/ci_run.go
bmeneguele: I'll take a closer look at this then. Because I've implemented the downstream pipeline in other commands (like `lab ci artifacts`) as a builtin feature, picking the downstream project/project through the current "upstream" pipeline information (API response). Letting the user to specify any project, regardless it being the downstream pipeline or not, is kinda weird considering how the remaining of the code and commands behaves.

I understand how useful it can be, I just think it's odd enough to break the consistency between commands. One thing I propose is: make this "--project" option strict to downstream pipelines by adding another option like done in `ci_artifacts` with `--follow`. What do you think?

bmeneguele:
@zaquestion I replied to your specific comments in the code before realizing you did a general comment in the PR.
My comments to your comments still applies though :)

bmeneguele:
ahh.. obviously I missed the `lab ci status`. Will check that as well and, possibly, open an issue to discuss it or to inform a possible conclusion (if we get to one here :smile:)

prarit:
@bmeneguele , this PR is still in draft after 20 days.  Is that intentional?  I've been holding off reviewing until it was "Ready".  Maybe that was a mistake on my part?

bmeneguele:
@prarit no, it's intentional. I need to come back to this code and rework. But unfortunately I had no time to rethink it yet.

bmeneguele:
To move this PR forward I'm going to remove the "deprecation" commit and move it to the next PR I'm preparing with the `--follow` logic.

bmeneguele:
@zampierilucas @prarit @zaquestion please, feel free to re-review this patch, like I said in the previous comment, the part that generated the whole discussion was removed.  I'll do that in the next PR.

zampierilucas (approved):

bmeneguele:
> downgrading my request for changes to a request for comment

Considering the above comment from @zaquestion and also the fact I'm going to open a new PR with the changes we discussed here so far, I'm going ahead and merge this PR.

PR #740, opened 2021-09-12
mr_create: Add -f option to open editor

### LINKED ISSUE #739: it would be nice if lab had an '-e' option which would open the editor
For example, a user executing the following command:

lab mr create --draft origin -F <filename> 

might want to specify an  option so that the file which is used as a template for the MR can be edited before the MR
is created.  For example:

lab mr create --draft origin -F <filename> -e

@prarit

prarit:
A request was made to add an option to open the editor when a file was
specified for the title and description.  Opening the editor would allow
project owners and maintainers to deploy MR templates that could be edited
by the MR submitter.

Add a --file-open_editor/-f option that opens the editor with the file
contents.

Closes #739

Suggested-by: David Arcari <darcari@redhat.com>

### DIFF
--- a/cmd/mr_create.go
+++ b/cmd/mr_create.go
@@ -34,6 +34,7 @@ var mrCreateCmd = &cobra.Command{
 		lab mr create my_remote --draft
 		lab mr create my_remote -F a_file.txt
 		lab mr create my_remote -F a_file.txt --force-linebreak
+		lab mr create my_remote -f a_file.txt
 		lab mr create my_remote -l bug -l confirmed
 		lab mr create my_remote -m "A title message"
 		lab mr create my_remote -m "A MR title" -m "A MR description"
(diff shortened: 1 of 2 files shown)


cmd/mr_create.go
```diff
@@ -54,7 +54,8 @@ func init() {
 	mrCreateCmd.Flags().BoolP("squash", "s", false, "squash commits when merging")
 	mrCreateCmd.Flags().Bool("allow-collaboration", false, "allow commits from other members")
 	mrCreateCmd.Flags().String("milestone", "", "set milestone by milestone title or ID")
-	mrCreateCmd.Flags().StringP("file", "F", "", "use the given file as the Description")
+	mrCreateCmd.Flags().StringP("file", "F", "", "use the given file as the Title and Description")
+	mrCreateCmd.Flags().StringP("file-open-editor", "f", "", "use the given file as the Title and Description and open the editor")
```
bmeneguele, line 58: What about `--file-edit` instead? Seems simpler and enough meaningful at the same time. What do you think?

cmd/mr_create.go
```diff
@@ -111,6 +112,22 @@ func runMRCreate(cmd *cobra.Command, args []string) {
...
 		log.Fatal(err)
 	}
+
+	ofilename, err := cmd.Flags().GetString("file-open-editor")
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	if ofilename != "" && filename != "" {
+		log.Fatalf("Cannot specify both -F and -o options.")
```
bmeneguele, line 122: I think you meant `-f` instead of `-o` correct? :))

cmd/mr_create.go
```diff
@@ -111,6 +112,22 @@ func runMRCreate(cmd *cobra.Command, args []string) {
...
+	}
+
+	openEditor := false
+	if ofilename != "" && filename == "" {
```
bmeneguele, line 126: If the previous `if` statement already checked if both options weren't used together, you don't need to check `filename == ""` here. If `filename != ""` it'll be caught in the previous conditional.

cmd/mr_create.go
prarit: Fixed in new version :)

cmd/mr_create.go
prarit: Fixed in new version :)

cmd/mr_create.go
bmeneguele: @prarit I think you forgot about this review of mine in your last update.

prarit: Oh I did!  Fixed in new version.

bmeneguele (approved):
LGTM

Issue #708, opened 2021-07-02
mr un/approve: If -F, un/approve completes even if file read failed


zampierilucas:
When using the -F, file, option, if the file read failed for any reason, the operation continues, ignoring the file comment.
In the example below, the user will unapprove the MR without submitting the message that he wanted.
```console
lzampier@trusty:~$ lab mr unapprove origin -F a_file_that_doest_exists
2021/07/02 10:53:46 ERROR: note_common.go:25: open a_file_that_doest_existis: no such file or directory
```
![image](https://user-images.githubusercontent.com/25491728/124285112-fffc5f80-db23-11eb-97e6-1ee30d0cc37e.png)

it is possible that the same problem may occur in other parts of Lab that provides the `-F` option, a global solution for that would be:
```diff
diff --git a/cmd/note_common.go b/cmd/note_common.go
index 789c3a9..08da617 100644
--- a/cmd/note_common.go
+++ b/cmd/note_common.go
@@ -4,6 +4,7 @@ import (
        "bytes"
        "fmt"
        "io/ioutil"
+       "os"
        "runtime"
        "strconv"
        "strings"
@@ -23,6 +24,7 @@ func createNote(rn string, isMR bool, idNum int, msgs []string, filename string,
                content, err := ioutil.ReadFile(filename)
                if err != nil {
                        log.Fatal(err)
+                       os.Exit(1)
                }
                body = string(content)
        } else {
```
Should Lab take care of that? if so let me know so I can do further testing and can submit a PR :)

zampierilucas:
@bmeneguele @prarit any input on this one? :)

prarit:
Huh, I thought the log.Fatal(err) would have stopped lab.  But yes, an os.Exit(1) there is fine.

zampierilucas:
@prarit on a quick search, I found that log.Fatal() __should__ call os.Exit(1).
I will investigate further this issue

bmeneguele:
Oh, yes. The `log.Fatal()` calls the `os.Exit()` it should not help to solve the problem.
The issue here might be the way `log.Fatal()` works, it doesn't close file descriptors nor running deferred code.
#729 might be related.

bmeneguele:
> Oh, yes. The `log.Fatal()` calls the `os.Exit()` it should not help to solve the problem.

Just to clarify: even though we have a custom log layer, our `log.Fatal()` still calls the default golang/logging `log.Fatal()`, so the behavior is kept.

[...]

PR #741, opened 2021-09-16
Review with quick actions

zampierilucas:
This is the second part of the change to unify the calls made when approving/commenting on an MR.

The new code removes the usage of the `merge_request/approve` API and instead uses the `/approve` and `/unapprove` notes API quick actions.
Fixes: #708

### DIFF
--- a/cmd/mr_approve.go
+++ b/cmd/mr_approve.go
@@ -63,7 +63,7 @@ var mrApproveCmd = &cobra.Command{
 			}
 		}
 
-		msgs = append(msgs, "\n  /approve")
+		msgs = append(msgs, "/approve")
 		createNote(rn, true, int(id), msgs, filename, linebreak, "", !note)
 
 		fmt.Printf("Merge Request !%d approved\n", id)
--- a/cmd/mr_approve.go
+++ b/cmd/mr_approve.go
@@ -61,6 +61,11 @@ var mrApproveCmd = &cobra.Command{
 			if err != nil {
 				log.Fatal(err)
 			}
+			if comment {
+				state := noteGetState(rn, true, int(id))
+				msg, _ := noteMsg(msgs, true, int(id), state, "", "")
+				msgs = append(msgs, msg)
+			}
 		}
 
 		msgs = append(msgs, "/approve")
--- a/cmd/mr_approve.go
+++ b/cmd/mr_approve.go
@@ -64,7 +64,7 @@ var mrApproveCmd = &cobra.Command{
 		}
 
 		msgs = append(msgs, "/approve")
-		createNote(rn, true, int(id), msgs, filename, linebreak, "", !note)
+		createNote(rn, true, int(id), msgs, filename, linebreak, "", note)
 
 		fmt.Printf("Merge Request !%d approved\n", id)
 	},
--- a/cmd/mr_approve.go
+++ b/cmd/mr_approve.go
@@ -3,12 +3,11 @@ package cmd
 import (
 	"fmt"
 	"github.com/MakeNowJust/heredoc/v2"
-	"os"
-
 	"github.com/rsteube/carapace"
 	"github.com/spf13/cobra"
 	"github.com/zaquestion/lab/internal/action"
 	lab "github.com/zaquestion/lab/internal/gitlab"
+	"os"
 )
 
 var mrApproveCmd = &cobra.Command{
@@ -28,9 +27,16 @@ var mrApproveCmd = &cobra.Command{
 			log.Fatal(err)
 		}
 
-		p, err := lab.FindProject(rn)
+		approvalConfig, err := lab.GetMRApprovalsConfiguration(rn, int(id))
 		if err != nil {
 			log.Fatal(err)
+		}
+
+		for _, approvers := range approvalConfig.ApprovedBy {
+			if approvers.User.Username == lab.User() {
+				fmt.Printf("Merge Request !%d already approved\n", id)
+				os.Exit(1)
+			}
 		}
 
 		comment, err := cmd.Flags().GetBool("with-comment")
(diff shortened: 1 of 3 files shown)


cmd/mr_approve.go
```diff
@@ -48,25 +54,17 @@ var mrApproveCmd = &cobra.Command{
...
-		if err != nil {
-			if err == lab.ErrStatusForbidden {
-				log.Fatal(err)
-			}
-			if err == lab.ErrActionRepeated {
-				fmt.Printf("Merge Request !%d already approved\n", id)
-				os.Exit(1)
-			}
-		}
+		msgs = append(msgs, "/approve")
```
bmeneguele, line 66: The `msg` var is not used when the user calls for a filename `-F`; shouldn't the quick action be added to the end of the file content when the user saves and closes it so he still get the "approve" action with the file content is used as the omment? (the same is valid for the unapprove code)

cmd/mr_unapprove.go
```diff
@@ -23,16 +23,32 @@ var mrUnapproveCmd = &cobra.Command{
 		lab mr unapprove upstream -m "A helpfull\nComment" --force-linebreak`),
 	PersistentPreRun: labPersistentPreRun,
 	Run: func(cmd *cobra.Command, args []string) {
+		var (
+			linebreak = false
+			canUnapprove = false
+		)
+
```
bmeneguele, line 30: Instead of creating this vars at the beginning, create them right before using them, like you did in the "approve" code, so can keep the code standard for both actions.

cmd/mr_unapprove.go
```diff
@@ -48,26 +64,17 @@ var mrUnapproveCmd = &cobra.Command{
...
+		if note {
+			linebreak, err = cmd.Flags().GetBool("force-linebreak")
 			if err != nil {
 				log.Fatal(err)
 			}
-
-			createNote(rn, true, int(id), msgs, filename, linebreak, "")
 		}
 
+		msgs = append(msgs, "/unapprove")
```
bmeneguele, line 75: The same situation I mentioned in the "approve" action.

cmd/note_common.go
```diff
@@ -283,7 +283,9 @@ func createNote(rn string, isMR bool, idNum int, msgs []string, filename string,
 	if err != nil {
 		log.Fatal(err)
 	}
-	fmt.Println(noteURL)
+	if !quickaction {
+		fmt.Println(noteURL)
+	}
```
bmeneguele, line 288: If a quick action is used with a comment the URL is still worthy showing, isn't it? It can be omitted only if no comments are given.

bmeneguele (changes requested):

cmd/note_common.go
zampierilucas: The reason why I've added that is that in the case of _just_ using the quick action, the URL returned is `https://gitlab.com/lzampier/kernel-test/merge_requests/0#note_000000000`. 
It sure seems like a bug on GitLab's side, but since they treat quick actions as a different kind of `message`, I do understand they having nothing to point at.

bmeneguele: Hmm, got it. What if instead of then omitting the URL completely you check its suffix for the invalid `#note_000000000`?
I know it's not documented and might be a bug on their side and, possibly, get fixed one day. But if we get a test at least for this case we can make sure we it failing as soon as the behavior changes on their side.

I just really like the URLs left in the stdout after the user performs an action that effectively created a valid URL he can click on :). But feel free to disagree.

zampierilucas: Wait, first things first, the URL is __not__ shown only when the quick action is sent only thing sent, like:
```shell
$ lab mr approve 
Merge Request !1 approved
```
If there is a comment, the output is:
```shell
$ lab mr approve -m "Im a comment"
https://gitlab.com/lzampier/kernel-test/merge_requests/1#note_683541416
Merge Request !1 approved
```
Knowing that, do you still want me to show that  '000*' URL? btw I'm all in for that since I can pretty much drop that `quickaction bool` var, and make the createNote function a bit cleaner :D

bmeneguele: Ok, I miss understood part of the code! That's my fault.
The `quickaction` variable name in `createNote()` mislead me because it's only `true` when the user wants to also comment (`note == true`). So my thought was: "wait, but the approval without comments are also being done with quick actions, so the `quickactions` var will be `true`there as well."

Maybe changing the name of the variable or adding some comments there would be helpful to avoid this kind of back and forth to understand what's going on, not a hard requirement, just IMO :), your call. But yes, the code is correct!

zampierilucas: Yeah, thank you for pointing that out, and I agree the code is not particularly clear on that point, I will submit a follow-up patch.

bmeneguele (approved):
Thanks @zampierilucas :)

zampierilucas:
I found an issue when using --with-comment

zampierilucas:
Submitted a new commit with a fix to --with-comment and everything should be working fine.

bmeneguele:
@zampierilucas I don't think the build issue you're facing is really related to your PR.
I just hit during the release testing: https://app.travis-ci.com/github/zaquestion/lab/builds/238658259

zampierilucas:
> @zampierilucas I don't think the build issue you're facing is really related to your PR. I just hit during the release testing: https://app.travis-ci.com/github/zaquestion/lab/builds/238658259

@bmeneguele GREAT!! I mean bad, really bad :p
but you're right `mr_note_test.go:102` is the same one that is breaking on my MR, maybe it has been broken for a while?
OTOH I've tested the command that sends `lab mr reply origin 1:689060304 -m "text` and It succeedss without a problem, so are we having some issue with our testing <-> gitlab.com relation again?

bmeneguele:
Possibly, but it shouldn't be panic'ing, but just dropping an error of any sort.
I'll try to take a look real quick to see if I can figure out and get the release out of the door.
However, if we can get it solved for the release, we can also get your PR also included before releasing :)

bmeneguele:
The issue is related to something on gitlab side. Two commands in the very same MR, but in different `notes` behaves differently with no apparent reason:
```
$ lab mr reply lab-testing 3953:599577526 -m "test"
https://gitlab.com/lab-testing/test/merge_requests/3953#note_689563923
$ lab mr reply lab-testing 3953:689488483 -m "test"
$ 
```
The first run returns what is expected, in the last one, it doesn't.
For "old" notes, it seems to work just fine, while for new ones, it doesn't.

Something changed in the API?
I'm going to dig a further.

bmeneguele:
For some really weird reason the discussions API is not returning new notes added to the MR, no matter if it's created using the Notes (where `individual_note = true`) or the Discussion API (where `individual_note = false`). 

I tested it using the API directly (through `curl`) so that's not an issue with `lab`, but definitely an issue on GitLab itself and a recent one, because the last PR that @prarit created was 4 days ago and everything was working just fine. I didn't find any issue open on gitlab.org though.

bmeneguele:
Well, the problem turned out to be a lack of an HTTP header in API responses for MRs/Issues with more than 10k comments, which happens in our testing repo.

@zampierilucas helped me to find the issue and the PR https://github.com/zaquestion/lab/pull/745 was created with the fix.

bmeneguele:
As soon as #745 is merged, this PR will need to be rebased.

prarit:
10k?  What have you been doing to my MRs? :smile:

zampierilucas:
@bmeneguele sure, I will rebase it as soon as 745 gets merged :D

> 10k? What have you been doing to my MRs? smile

@prarit GL considers 'entries' as any historical data on the MR, not only 'comments', for example, 'X person approved the MR' is also an entry, That's why we were able to beet GitLab 10k restriction that fast :p

prarit:
Entries?  I'm just going to assume that somewhere there is a character-by-character review of one of my changesets.

bmeneguele:
@zampierilucas #745 is now merged, please go ahead and rebase this one here so we can make sure it's also fine :)

bmeneguele (approved):

PR #745, opened 2021-09-30
gitlab: handle API response without X-Total(-Pages) HTTP header

bmeneguele:
In cases where the number of comments in a certain Issue/MR is greater than
10k the HTTP header for the API responses lack the X-Total and X-Total-Pages
metadata [1] used by lab to control the response pagination. Even though it
should not happen in real scenarios, it's breaking our tests where some MRs
have more than 10k comments. The solution adopted was to use the X-Next-Page
header whenever the X-Total(-Page) is not found.

[1] https://docs.gitlab.com/ee/user/gitlab_com/index.html#pagination-response-headers

### DIFF
--- a/internal/gitlab/gitlab.go
+++ b/internal/gitlab/gitlab.go
@@ -421,10 +421,12 @@ func MRList(project string, opts gitlab.ListProjectMergeRequestsOptions, n int) 
 	if err != nil {
 		return nil, err
 	}
-	if resp.CurrentPage == resp.TotalPages {
+
+	var ok bool
+	if opts.Page, ok = hasNextPage(resp); !ok {
 		return list, nil
 	}
-	opts.Page = resp.NextPage
+
 	for len(list) < n || n == -1 {
 		if n != -1 {
 			opts.PerPage = n - len(list)
@@ -433,12 +435,13 @@ func MRList(project string, opts gitlab.ListProjectMergeRequestsOptions, n int) 
 		if err != nil {
 			return nil, err
 		}
-		opts.Page = resp.NextPage
 		list = append(list, mrs...)
-		if resp.CurrentPage == resp.TotalPages {
+
+		if opts.Page, ok = hasNextPage(resp); !ok {
 			break
 		}
 	}
+
 	return list, nil
 }
 
@@ -501,13 +504,12 @@ func MRListDiscussions(project string, mrNum int) ([]*gitlab.Discussion, error) 
 		// ... and add them to our collection of discussions
 		discussions = append(discussions, d...)
 
-		// if we've seen all the pages, then we can break here
-		if resp.CurrentPage >= resp.TotalPages {
+		// if we've seen all the pages, then we can break here.
+		// otherwise, update the page number to get the next page.
+		var ok bool
+		if opt.Page, ok = hasNextPage(resp); !ok {
 			break
 		}
-
-		// otherwise, update the page number to get the next page.
-		opt.Page = resp.NextPage
 	}
 
 	return discussions, nil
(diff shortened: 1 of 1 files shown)


internal/gitlab/gitlab.go
```diff
@@ -1900,3 +1906,18 @@ func CreateMergeRequestCommitDiscussion(project string, mrID int, sha string, ne
...
+
+// hasNextPage is a pagination helper to handle the case where the API
+// response comes without the HTTP X-Total(-Page) header. Reference:
+// https://docs.gitlab.com/ee/user/gitlab_com/index.html#pagination-response-headers
+func hasNextPage(resp *gitlab.Response) (int, bool) {
+	if resp.TotalPages != 0 {
+		if resp.CurrentPage >= resp.TotalPages {
+			return 0, false
+		}
+	} else if resp.CurrentPage >= resp.NextPage {
```
zampierilucas, line 1918: What happens in the case of, let's say, I have an API with 10 pages, the ninth iteration will return return `resp.NextPage` as 9, and on the tenth? I'm assuming NextPage will be zero by then so we return 0 and end up not retrieving page 10?

prarit, line 1918: So suppose there are 100 pages ... what will the 99th resp.NextPage return?  0?

prarit: I see @zampierilucas asked the same question.  I'll resolve this and you can answer it once.

internal/gitlab/gitlab.go
```diff
@@ -1900,3 +1906,18 @@ func CreateMergeRequestCommitDiscussion(project string, mrID int, sha string, ne
...
 }
+
+// hasNextPage is a pagination helper to handle the case where the API
+// response comes without the HTTP X-Total(-Page) header. Reference:
+// https://docs.gitlab.com/ee/user/gitlab_com/index.html#pagination-response-headers
+func hasNextPage(resp *gitlab.Response) (int, bool) {
+	if resp.TotalPages != 0 {
+		if resp.CurrentPage >= resp.TotalPages {
+			return 0, false
+		}
```
zampierilucas, line 1917: I don't see the point of keeping this code, since we are returning NextPage anyway.

internal/gitlab/gitlab.go
bmeneguele: I'm not sure if I follow your concern. The X-Page starts at `1`, so in the 9th iteration `X-Page: 9` and `X-Next-Page: 10`, so we request the `X-Page: 10`, which is the last one. With that, `resp.CurrentPage == resp.NextPage`. In theory, `CurrentPage` will never be `>` than `NextPage`.

bmeneguele: `X-Page` (`CurrentPage`) starts at 1, so when `CurrentPage == 100` the `X-Next-Page` will be empty in the HTTP header and the go-gitlab lib will make it `0`.

prarit: Ah okay.  I think both @zampierilucas were confused by the numbering.  That make sense now.

internal/gitlab/gitlab.go
bmeneguele: That's a good point. Considering that only `X-Total` and `X-Total-Pages` might not be present in the header, I think we can just not use it at all. I'll remove it and send a new version.

internal/gitlab/gitlab.go
zampierilucas: Yeah, Thanks @bmeneguele

prarit (approved):

PR #630, opened 2021-03-10
Git command warning

prarit:
A few users have been asking about lab's purpose relative to git.  The current description of the tool "A Git Wrapper for GitLab" is not accurate and should be updated to reflect that lab is a command line utility for GitLab.  

Also, output a warning for git wrapper users -- we know this code is going away in the future and we should be encouraging people to not write scripts that depend on lab's git wrapper.

Signed-off-by: Prarit Bhargava <prarit@redhat.com>

PR #770, opened 2021-12-09
Remove documentation about using lab as a replacement for git

JohnVillalovos:
Based on #630, it appears that the project no longer wants people to
use lab as a replacement for the git command. So we shouldn't
recommend to people that the do `alias git=lab`.

Update README.md to remove this recommendation.


bmeneg (approved):
Ty! We should've done that already.

PR #771, opened 2021-12-10
cmd/note-common.go: Fix context comments on diffs

prarit:
@krobelus noticed that it wasnt possible to comment on the context portion
of a diff.  For example, when commenting on line 6 of

| newfile: Makefile oldfile: Makefile
|        @@ -5,6 +5,8 @@ SUBLEVEL = 0
|  5   5  EXTRAVERSION =
|  6   6  NAME = Merciless Moray
|  7   7
|      8 +# space
|      9 +
|  8  10  #
|  9  11  # DRM backport version
| 10  12  #

it was not possible to comment on line 6 and lab would output an error.
This is fixed by reworking the code to detect context diffs, however, this
only appears to work on code shown before the difflines.  There are
several open GitLab API issues that report errors when commenting on Merge
Request commits and it is likely that one of these is causing the error.

ex) https://gitlab.com/gitlab-org/gitlab-foss/-/issues/28599

Rework the code to allow for commenting on the context of diffs.

### DIFF
--- a/cmd/note_common.go
+++ b/cmd/note_common.go
@@ -94,11 +94,9 @@ func noteRunFn(cmd *cobra.Command, args []string) {
 	createNote(rn, isMR, int(idNum), msgs, filename, linebreak, commit, true)
 }
 
-func createCommitNote(rn string, mrID int, sha string, newFile string, oldFile string, oldline int, newline int, comment string, block bool) {
-	linetype := "old"
+func createCommitNote(rn string, mrID int, sha string, newFile string, oldFile string, linetype string, oldline int, newline int, comment string, block bool) {
 	line := oldline
 	if oldline == -1 {
-		linetype = "new"
 		line = newline
 	}
 
@@ -145,63 +143,106 @@ func createCommitComments(project string, mrID int, commit string, body string, 
 	// tracking information (new line & old line number pairs,
 	// and file information)
 	scanner := bufio.NewScanner(strings.NewReader(body))
+	lastDiffLine := ""
+	comments := ""
 	newfile := ""
 	oldfile := ""
-	oldLineNum := -1
-	newLineNum := -1
-	comments := ""
+	diffCut := 1
+
 	for scanner.Scan() {
-		if strings.HasPrefix(scanner.Text(), "| newfile:") {
-			if comments != "" {
-				createCommitNote(project, mrID, commit, newfile, oldfile, oldLineNum, newLineNum, comments, block)
-				comments = ""
+		line := scanner.Text()
+
+		if !strings.HasPrefix(line, "| ") {
+			comments += "\n" + line
+			continue
+		}
+
+		if comments != "" && lastDiffLine == "" {
+			log.Fatal("Cannot comment on first line of commit (context unknown).")
+		}
+
+		if comments != "" {
+			linetype := ""
+			oldLineNum := -1
+			newLineNum := -1
+
+			// parse lastDiffLine
+			f := strings.Fields(strings.TrimSpace(lastDiffLine))
+
+			// The output can be, for example:
+			// | # # [no +/-] < context comment
+			// | # #    +     < newline comment
+			// | # #    -     < oldline comment
+			// | #      -     < oldline comment
+			// |   #    +     < newline comment
+
+			// f[0] is always a "|"
+			// f[1] will always be a number
+			val1, _ := strconv.Atoi(f[1])
+			val2, err := strconv.Atoi(f[2])
+
+			if err == nil { // f[2] is a number
+				if len(f) <= 3 { // f[3] does not exist
+					oldLineNum = val1
+					newLineNum = val2
+					linetype = "context"
+				} else {
+					newLineNum = val2
+					switch {
+					case strings.HasPrefix(f[3], "+"):
+						linetype = "new"
+					case strings.HasPrefix(f[3], "-"):
+						linetype = "old"
+					default:
+						linetype = "context"
+					}
+				}
+			} else { // f[2] is not a number
+				switch {
+				case strings.HasPrefix(f[2], "+"):
+					newLineNum = val1
+					linetype = "new"
+				case strings.HasPrefix(f[2], "-"):
+					oldLineNum = val1
+					linetype = "old"
+				default:
+					panic("unknown string in diff")
+				}
 			}
+
+			createCommitNote(project, mrID, commit, newfile, oldfile, linetype, oldLineNum, newLineNum, comments, block)
+			comments = ""
+		}
+
+		f := strings.Fields(strings.TrimSpace(line))
+		if f[1] == "@@" {
+			// In GitLab diff output, the leading "@" symbol indicates where
+			// the metadata ends.  This is true even if passing over a digit
+			// boundary (ie going from line 99 to 100).  This location can
+			// be used to truncate the lines to only include the metadata.
+			diffCut = strings.Index(line, "@") + 1
+			lastDiffLine = ""
+			continue
+		}
+
+		if f[1] == "newfile:" {
 			// read filename
-			f := strings.Split(scanner.Text(), " ")
+			f := strings.Split(line, " ")
 			newfile = f[2]
 			if len(f) < 5 {
 				oldfile = ""
 			} else {
 				oldfile = f[4]
 			}
-		} else if strings.HasPrefix(scanner.Text(), "|") {
-			if comments != "" {
-				createCommitNote(project, mrID, commit, newfile, oldfile, oldLineNum, newLineNum, comments, block)
-				comments = ""
-			}
-			// read line numbers
-			fs := strings.Split(scanner.Text(), " ")
+			continue
+		}
 
-			oldLineNum = -1
-			newLineNum = -1
-			for _, f := range fs {
-				if f == "" || f == "|" || f == "@@" {
-					continue
-				}
-				val, err := strconv.Atoi(f)
-				if err != nil {
-					// NaN
-					if strings.HasPrefix(f, "+") {
-						newLineNum = oldLineNum
-						oldLineNum = -1
-					}
-					break
-				} else {
-					// Number
-					if oldLineNum == -1 {
-						oldLineNum = val
-					} else {
-						newLineNum = val
-						break
-					}
-				}
-			}
+		if len(line) > diffCut {
+			lastDiffLine = line[0:diffCut]
 		} else {
-			// this is a comment (combine for a filename)
-			comments = comments + "\n" + scanner.Text()
+			lastDiffLine = line
 		}
 	}
-
 }
 
 func noteGetState(rn string, isMR bool, idNum int) (state string) {
--- a/internal/gitlab/gitlab.go
+++ b/internal/gitlab/gitlab.go
@@ -1718,12 +1718,18 @@ func CreateCommitComment(projID string, sha string, newFile string, oldFile stri
 		PositionType: "text",
 	}
 
-	if linetype == "old" {
-		position.OldPath = oldFile
-		position.OldLine = line
-	} else {
+	switch linetype {
+	case "new":
 		position.NewPath = newFile
 		position.NewLine = line
+	case "old":
+		position.OldPath = oldFile
+		position.OldLine = line
+	case "context":
+		position.NewPath = newFile
+		position.NewLine = line
+		position.OldPath = oldFile
+		position.OldLine = line
 	}
 
 	opt := &gitlab.CreateCommitDiscussionOptions{
@@ -1760,9 +1766,13 @@ func CreateMergeRequestCommitDiscussion(projID string, id int, sha string, newFi
 		PositionType: "text",
 	}
 
-	if linetype == "new" {
+	switch linetype {
+	case "new":
 		position.NewLine = line
-	} else {
+	case "old":
+		position.OldLine = line
+	case "context":
+		position.NewLine = line
 		position.OldLine = line
 	}
 


prarit:
@krobelus @bmeneg @zampierilucas

prarit:
Frustratingly I cannot get the code to work for context lines at the end of the diff :/  I've tried everything I could think of, including adding linePosition entries and forcing the line_code.  I still get the same error.

cmd/note_common.go
```diff
@@ -94,11 +94,9 @@ func noteRunFn(cmd *cobra.Command, args []string) {
 	createNote(rn, isMR, int(idNum), msgs, filename, linebreak, commit, true)
```
krobelus, line 94: I appreciate the nice commit message but please don't @ people in commit messages,
because github will send a notification email whenever someone pushes that commit
to a public fork. I haven't checked if they it's possible to turn that off.

The essence of the change is that we pass *both* line numbers for context lines, to please gitlab.

cmd/note_common.go
```diff
@@ -166,42 +165,64 @@ func createCommitComments(project string, mrID int, commit string, body string,
 			}
 		} else if strings.HasPrefix(scanner.Text(), "|") {
 			if comments != "" {
-				createCommitNote(project, mrID, commit, newfile, oldfile, oldLineNum, newLineNum, comments, block)
+				createCommitNote(project, mrID, commit, newfile, oldfile, linetype, oldLineNum, newLineNum, comments, block)
```
krobelus, line 168: > Frustratingly I cannot get the code to work for context lines at the end of the diff :/  I've tried everything I could think of, including adding linePosition entries and forcing the line_code.  I still get the same error.

Anyway here is a minimal fix. I had to insert a

```go
	fmt.Printf("%#v\n", position)
```

which showed that we send the wrong "old" line for context lines.
Would be nice if `lab` could print `curl` commands, then it'd be easier to check with other tools and the API docs.

```diff
diff --git a/cmd/note_common.go b/cmd/note_common.go
index 6f6bd14..4bf44e9 100644
--- a/cmd/note_common.go
+++ b/cmd/note_common.go
@@ -94,14 +94,9 @@ func noteRunFn(cmd *cobra.Command, args []string) {
 	createNote(rn, isMR, int(idNum), msgs, filename, linebreak, commit, true)
 }
 
-func createCommitNote(rn string, mrID int, sha string, newFile string, oldFile string, linetype string, oldline int, newline int, comment string, block bool) {
-	line := oldline
-	if oldline == -1 {
-		line = newline
-	}
-
+func createCommitNote(rn string, mrID int, sha string, newFile string, oldFile string, linetype string, oldLine int, newLine int, comment string, block bool) {
 	if block {
-		webURL, err := lab.CreateMergeRequestCommitDiscussion(rn, mrID, sha, newFile, oldFile, line, linetype, comment)
+		webURL, err := lab.CreateMergeRequestCommitDiscussion(rn, mrID, sha, newFile, oldFile, oldLine, newLine, linetype, comment)
 		if err != nil {
 			log.Fatal(err)
 		}
@@ -109,7 +104,7 @@ func createCommitNote(rn string, mrID int, sha string, newFile string, oldFile s
 		return
 	}
 
-	webURL, err := lab.CreateCommitComment(rn, sha, newFile, oldFile, line, linetype, comment)
+	webURL, err := lab.CreateCommitComment(rn, sha, newFile, oldFile, oldLine, newLine, linetype, comment)
 	if err != nil {
 		log.Fatal(err)
 	}
@@ -190,6 +185,7 @@ func createCommitComments(project string, mrID int, commit string, body string,
 			val2, err := strconv.Atoi(f[2])
 			if err == nil { // f[2] is a number
 				if len(f) > 3 { // f[3] exists
+					oldLineNum = val1
 					newLineNum = val2
 					// f[3] starts with a +
 					if strings.HasPrefix(f[3], "+") {
diff --git a/internal/gitlab/gitlab.go b/internal/gitlab/gitlab.go
index 3793c3f..d6696bc 100644
--- a/internal/gitlab/gitlab.go
+++ b/internal/gitlab/gitlab.go
@@ -1695,7 +1695,7 @@ func GetCommitDiff(projID string, sha string) ([]*gitlab.Diff, error) {
 	return diffs, nil
 }
 
-func CreateCommitComment(projID string, sha string, newFile string, oldFile string, line int, linetype string, comment string) (string, error) {
+func CreateCommitComment(projID string, sha string, newFile string, oldFile string, oldLine int, newLine int, linetype string, comment string) (string, error) {
 	// Ideally want to use lab.Commits.PostCommitComment, however,
 	// that API only support comments on linetype=new.
 	//
@@ -1721,15 +1721,15 @@ func CreateCommitComment(projID string, sha string, newFile string, oldFile stri
 	switch linetype {
 	case "new":
 		position.NewPath = newFile
-		position.NewLine = line
+		position.NewLine = newLine
 	case "old":
 		position.OldPath = oldFile
-		position.OldLine = line
+		position.OldLine = oldLine
 	case "context":
 		position.NewPath = newFile
-		position.NewLine = line
+		position.NewLine = newLine
 		position.OldPath = oldFile
-		position.OldLine = line
+		position.OldLine = oldLine
 	}
 
 	opt := &gitlab.CreateCommitDiscussionOptions{
@@ -1745,7 +1745,7 @@ func CreateCommitComment(projID string, sha string, newFile string, oldFile stri
 	return fmt.Sprintf("%s#note_%d", commitInfo.WebURL, commitDiscussion.Notes[0].ID), nil
 }
 
-func CreateMergeRequestCommitDiscussion(projID string, id int, sha string, newFile string, oldFile string, line int, linetype string, comment string) (string, error) {
+func CreateMergeRequestCommitDiscussion(projID string, id int, sha string, newFile string, oldFile string, oldLine int, newLine int, linetype string, comment string) (string, error) {
 	commitInfo, err := GetCommit(projID, sha)
 	if err != nil {
 		fmt.Printf("Could not get diff for commit %s.\n", sha)
@@ -1768,12 +1768,12 @@ func CreateMergeRequestCommitDiscussion(projID string, id int, sha string, newFi
 
 	switch linetype {
 	case "new":
-		position.NewLine = line
+		position.NewLine = newLine
 	case "old":
-		position.OldLine = line
+		position.OldLine = oldLine
 	case "context":
-		position.NewLine = line
-		position.OldLine = line
+		position.NewLine = newLine
+		position.OldLine = oldLine
 	}
 
 	opt := &gitlab.CreateMergeRequestDiscussionOptions{
```

cmd/note_common.go
```diff
@@ -166,42 +165,64 @@ func createCommitComments(project string, mrID int, commit string, body string,
...
 				comments = ""
 			}
 			// read line numbers
-			fs := strings.Split(scanner.Text(), " ")
+			f := strings.Split(strings.Join(strings.Fields(strings.TrimSpace(scanner.Text())), " "), " ")
```
krobelus, line 172: according to their documentation, `strings.Split(strings.Join(x, " "), " ")` should be a no-op?

cmd/note_common.go
```diff
@@ -166,42 +165,64 @@ func createCommitComments(project string, mrID int, commit string, body string,
...
+
+			// f[0] is always a "|"
+			// f[1] will always be a number
+			val1, _ := strconv.Atoi(f[1])
+			val2, err := strconv.Atoi(f[2])
+			if err == nil { // f[2] is a number
+				if len(f) > 3 { // f[3] exists
+					newLineNum = val2
+					// f[3] starts with a +
+					if strings.HasPrefix(f[3], "+") {
```
krobelus, line 195: the last two comments are too obvious IMO :(

cmd/note_common.go
```diff
@@ -166,42 +165,64 @@ func createCommitComments(project string, mrID int, commit string, body string,
...
-						oldLineNum = val
-					} else {
-						newLineNum = val
-						break
+					// f[3] starts with a -
+					if strings.HasPrefix(f[3], "-") {
+						linetype = "old"
 					}
+					// f[3] does not start with a +/-
+					linetype = "context"
```
krobelus, line 203: this stomps the assignments above, maybe use else-if or switch?

cmd/note_common.go
```diff
@@ -166,42 +165,64 @@ func createCommitComments(project string, mrID int, commit string, body string,
...
+				} else { // f[3] does not exist
+					oldLineNum = val1
+					newLineNum = val2
+					linetype = "context"
+				}
+			} else { // f[2] is not a number
```
krobelus, line 209: I personally don't like all those nested conditionals.

Instead of

```go
if x {
	doSomething()
} else { // not x
	doSomethingElse()
	doSomethingElse()
	doSomethingElse()
}
```

I'd use early return/continue

```go
if x {
	doSomething()
	continue
}
doSomethingElse()
doSomethingElse()
doSomethingElse()
```

though I haven't actually tried it here, feel free to ignore

cmd/note_common.go
```diff
@@ -166,42 +165,64 @@ func createCommitComments(project string, mrID int, commit string, body string,
...
+					// f[3] starts with a -
+					if strings.HasPrefix(f[3], "-") {
+						linetype = "old"
 					}
+					// f[3] does not start with a +/-
+					linetype = "context"
+				} else { // f[3] does not exist
+					oldLineNum = val1
+					newLineNum = val2
+					linetype = "context"
```
krobelus, line 207: this "else" case is shorter than the "then" case so I'd maybe switch their order. Checking the error case first also means everything below can assume there were no errors :innocent:

cmd/note_common.go
krobelus, line 168: The fact that we duplicate this means it may be better to restructure this loop. Currently we have:

```go
for scanner.Scan() {
	if strings.HasPrefix(scanner.Text(), "| newfile:") {
		createCommitNote(project, mrID, commit, newfile, oldfile, linetype, oldLineNum, newLineNum, comments, block)
		// short body
	} else if strings.HasPrefix(scanner.Text(), "|") {
		createCommitNote(project, mrID, commit, newfile, oldfile, linetype, oldLineNum, newLineNum, comments, block)
		// long body
	} else {
		comments += "\n" + scanner.Text() // very short body
	}
}
```

I'd rather read this
(extracting scanner.Text() because it allocates on every call)

```go
for scanner.Scan() {
	line := scanner.Text()
	if !strings.HasPrefix(line, "|") {
		comments += "\n" + line
		continue
	}

	createCommitNote(project, mrID, commit, newfile, oldfile, linetype, oldLineNum, newLineNum, comments, block)
	
	if strings.HasPrefix(line, "| newfile:") {
		// short body
		continue
	}

	// long body
}
```

**However** since commenting on the very last context line of the last file in a diff doesn't work because there is no following "|" line, we should probably restructure this to just compute a list of comments, and submit all of them once we're done.

cmd/note_common.go
```diff
@@ -166,42 +165,64 @@ func createCommitComments(project string, mrID int, commit string, body string,
...
+				// f[2] starts with a +
+				if strings.HasPrefix(f[2], "+") {
+					newLineNum = val1
+					linetype = "new"
+				}
+				// f[2] starts with a -
+				if strings.HasPrefix(f[2], "-") {
+					oldLineNum = val1
+					linetype = "old"
 				}
```
krobelus, line 219: maybe error if we get anything but + or -

cmd/note_common.go
```diff
@@ -166,42 +165,64 @@ func createCommitComments(project string, mrID int, commit string, body string,
...
 			if comments != "" {
-				createCommitNote(project, mrID, commit, newfile, oldfile, oldLineNum, newLineNum, comments, block)
+				createCommitNote(project, mrID, commit, newfile, oldfile, linetype, oldLineNum, newLineNum, comments, block)
 				comments = ""
 			}
 			// read line numbers
-			fs := strings.Split(scanner.Text(), " ")
+			f := strings.Split(strings.Join(strings.Fields(strings.TrimSpace(scanner.Text())), " "), " ")
+			if f[1] == "@@" {
+				continue;
```
krobelus, line 174: gofmt should remove the ;
I see that we are still using travis, which is basically dead AFAIK. We should either move to something like GH actions or move to gitlab straight away

cmd/note_common.go
prarit: Oops, I had no idea.  Fixed.

cmd/note_common.go
prarit: Yeah, that's what I've been doing.  But even when I do set a "valid" position at the end of the commit, the command to comment returns an error:
2021/12/10 19:11:31 ERROR: note_common.go:114: POST https://gitlab.com/api/v4/projects/redhat/rhel/src/kernel/kernel-test/repository/commits/abf523e2e10da39b15f522378838bf4d39604e15/discussions: 400 {message: 400 Bad request - Note {:line_code=>["can't be blank", "must be a valid line code"]}}

prarit: I'll look into seeing if go-gitlab has a debug option.  IIRC it does but I've never used it.

cmd/note_common.go
prarit: Oh, that seems to work.  I clearly overthought that :) .  Fixed.

cmd/note_common.go
prarit: Yeah, it's more for me to read through TBH ;).  I'd prefer to keep them in place for now.

cmd/note_common.go
prarit: Thanks.  Fixed.

cmd/note_common.go
prarit: Yeah, I've spent some time staring at this code and thinking about ways to reorganize it.  I think a complete rewrite might be in order ;) but that's something for another PR.

cmd/note_common.go
prarit: Oh good one.  Fixed.

cmd/note_common.go
prarit: Ah, good suggestion.  I've incorporated this change in an additional cleanup commit.

cmd/note_common.go
prarit: I was thinking about that and wondering if it was ever possible.  I've thrown in a panic()

cmd/note_common.go
prarit: Damned c coding style.  It's my fingers :) I can't get away from it.  Fixed.

cmd/note_common.go
krobelus: I didn't get any errors with this diff applied, do you have a reproducible example?

internal/gitlab/gitlab.go
```diff
@@ -1751,18 +1757,42 @@ func CreateMergeRequestCommitDiscussion(projID string, id int, sha string, newFi
 		return "", err
 	}
 
+	sr := gitlab.LinePosition{
+		LineCode: sha + "_8_10",
```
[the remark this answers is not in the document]
prarit: Woops. I left in some ugly debug code.  I've removed it.  My apologies for that.

cmd/note_common.go
```diff
@@ -164,44 +170,65 @@
...
-				}
+			// f[3] exists
+			newLineNum = val2
+			// f[3] starts with a +
+			if strings.HasPrefix(f[3], "+") {
+				linetype = "new"
+				continue
+			} else if strings.HasPrefix(f[3], "-") { // f[3] starts with a -
+				linetype = "old"
+				continue
```
krobelus, line 211: I think we need to set `oldLineNum` in this case.
Also in the "context" case (in addition to newLineNum). I believe we can just add a `oldLineNum = val1` next to `newLineNum = val1`

LGTM otherwise from a quick look.

cmd/note_common.go
```diff
@@ -164,44 +170,65 @@
...
+		f := strings.Fields(strings.TrimSpace(scanner.Text()))
+		if f[1] == "@@" {
+			continue
+		}
+
+		oldLineNum = -1
+		newLineNum = -1
+
+		// The output can be, for example:
+		// | # # [no +/-] < context comment
```
krobelus, line 186: there's another bug: if a context line starts with - or +, we interpret it as - or + line:

	| newfile: README.md oldfile: README.md
	|      @@ -1,5 +1,6 @@
	| 1  1  - context line
	|    2 +- added line 

Probably shouldn't be using strings.Fields but a regex like `^\|(?: +(\d+)){1,2} ([-+ ])`

cmd/note_common.go
prarit: So something like this?  I think this should work.

```
diff --git a/cmd/note_common.go b/cmd/note_common.go
index be02a45afb21..17fd5386c86b 100644
--- a/cmd/note_common.go
+++ b/cmd/note_common.go
@@ -201,6 +201,7 @@ func createCommitComments(project string, mrID int, commit string, body string,
                                continue
                        }
                        // f[3] exists
+                       oldLineNum = val1
                        newLineNum = val2
                        // f[3] starts with a +
                        if strings.HasPrefix(f[3], "+") {

```

krobelus: Yes but seeing that we do the same in the if-statement above, we can just move these two lines before the if statement.

cmd/note_common.go
prarit: Good point.  I'm going to fix this up in a follow-up commit.  I think the same bug was there before this change.

cmd/note_common.go
prarit: Hmm, it doesn't appear to work that way.  If a new line comment is made the old line must be not set AFAICT.

krobelus: > Hmm, it doesn't appear to work that way. If a new line comment is made the old line must be not set AFAICT.

you're right, my bad

prarit:
Hey @krobelus , I've uploaded a new version that takes all of your concerns into account and also fixes a bug where users should not be able to comment on the first line of the patch.

prarit:
@bmeneg care to review?

cmd/note_common.go
```diff
@@ -145,17 +143,83 @@ func createCommitComments(project string, mrID int, commit string, body string,
...
+
+		f := strings.Fields(strings.TrimSpace(line))
+		if f[1] == "@@" {
+			lastDiffLine = ""
+			continue
+		}
+
+		if f[1] == "newfile:" {
 			// read filename
 			f := strings.Split(scanner.Text(), " ")
```
bmeneg, line 224: Just minor nit: since we already have `line` at this point we wouldn't need to call for `scanner.Text()`.

cmd/note_common.go
```diff
@@ -164,44 +228,15 @@ func createCommitComments(project string, mrID int, commit string, body string,
...
-			// this is a comment (combine for a filename)
-			comments = comments + "\n" + scanner.Text()
+			continue
 		}
-	}
 
+		// The first character of text is preceded by a tab.  This is used in other
+		// utilities to differentiate between the diff metadata and codetext from the diff
+		// itself.
+		tabline := strings.Split(line, "\t")
```
bmeneg, line 237: I didn't get the idea of this `\t` preceding the first char. Would you mind explaining it a bit more? (not sure if I missed something in the previous discussions).

bmeneg (commented):
Overall seems fine, just a small nitpick and a doubt about a code.
No real need for changing anything yet.

cmd/note_common.go
prarit: Fixed, and added a better algorithm to detect where the metadata is.  I now am using the location of the first "@" symbol in the text.  This always lines up with the metadata in the GitLab diff.  For example,
```

|          @@ -98,6 +98,8 @@ else
|  98   98    Q = @
|  99   99  endif
| 100  100  
|      101 +#somecodehere
|      102 +
| 101  103  # If the user is running make -s (silent mode), suppress echoing of
| 102  104  # commands
| 103  105  
```
so the location of the first "@" can be used to truncate the lines to find the metadata.

bmeneg: Ahh , I see it now.
Thanks @prarit!

bmeneg (approved):

PR #779, opened 2022-02-01
mr checkout: add --track-default-remote flag (closes #774)

### LINKED ISSUE #774: `mr checkout`: specify the tracking remote
I don't use the "1 remote/author" workflow at work; instead we all push branches to a single, shared `origin` where only some of us have permission to merge to `main`. As written, `mr checkout -t <id>` will create/update a remote named after the MR author. In my case, this ends up w/ 2 remotes both pointing to the same URL, which doesn't feel helpful to me. I have hacked around this locally, but before I go making it pretty I'd like to know if anyone else wants this. 😛 

My existing hack is just an add'l flag `--track-default-remote` that tells `mr checkout` use `targetRemote` instead of creating a new remote based on `mr.Author.Username`.

Any interest?

BTW I looked into some way to support something like this:
- `--tracking-remote` not specified on command line: use `mr.Author.Username` (current behavior)
- `--tracking-remote` specified but w/ no value: use default remote
- `--tracking-remote=<remoteName>`: use remote named `remoteName`

I got stuck on this b/c I couldn't (quickly) figure out how to coax pflags into working w/ some sort of flag that has different values if it's not specified, vs if it's passed but w/o a value vs w/ a value. (Or even if such a thing is possible.) I eventually just settled for the default remote being fine for me. 😄

claytonrcarter:
Hi again, this adds a new `--track-default-remote` flag to `mr checkout`, as discussed in #774. My Go is pretty rusty (well, it's never *not* been rusty) so I don't know if I've stuck to idioms, etc. Happy to make updates as requested, of course.

If specified, `--track-default-remote` implies `--track` but also alters it so that instead of tracking a remote named after the MR author, it will use either the remote named in `remote.pushDefault` or the value of `targetRemote` (which is just the value of `defaultRemote`). 

~~I did not add a test, but this has been working for me daily for weeks now. I did play around w/ a test but got stymied along the way. I would be happy to add a test if someone can provide me w/ some guidance.~~ **edit:** I added a test that looks at what the new branch is tracking, and also tweaked an existing test to confirm that it is also tracking the correct remote/branch.

Thank you for considering this!

### DIFF
--- a/cmd/mr_checkout_test.go
+++ b/cmd/mr_checkout_test.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"os/exec"
+	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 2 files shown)


prarit:
@zampierilucas can you work with @claytonrcarter on helping him design a test?

claytonrcarter:
Hi again. I figured out a testing strategy and added a test. I've also squashed it all back into 1 commit.

cmd/mr_checkout_test.go
```diff
@@ -88,6 +89,42 @@ func Test_mrCheckoutCmd_track(t *testing.T) {
...
 	require.Contains(t, remotes, "zaquestion	git@gitlab.com:zaquestion/test.git")
+
+	// confirm that new branch tracks zaquestion/mrtest
+	cmd = exec.Command("git", "rev-parse", "--symbolic-full-name", "--abbrev-ref", "mrtest_track@{upstream}")
+	cmd.Dir = repo
+	gitOut, err = cmd.CombinedOutput()
+	if err != nil {
+		t.Fatal(err)
+	}
+	require.Equal(t, "zaquestion/mrtest", strings.TrimSpace(string(gitOut)))
```
claytonrcarter, line 100: This assertion is meant to confirm that the new branch is set to track the upstream `zaquestion/mrtest`. My understanding is that the existing test looks for various circumstantial signs that this has been set up correctly, but doesn't actually confirm the specific upstream relation. Happy to revert this if you feel it's not necessary. Thanks!

bmeneg:
@claytonrcarter it seems we can have a simpler solution for this case without adding a new flag (check issue #774).
I can come up with something for you to test.

claytonrcarter:
Thanks! Closing in favor of #786

PR #783, opened 2022-02-03
Added discussion and ctivity filters

zampierilucas:
Added two new command line options that allow to filter by discussion or
activities in both MRs and Issues, keeping the default being showing both.

Talking to Bruno, we couldn't find a good naming for `--discussions-only` and `--activities-only`. 
We could shorten those to `--discussion` and `--activity` but I wonder if that might be confusing with the usage `--comment`.
Maybe we could use `--only=discussion` or `--only=activity`, I'm not particularly a fan of that notation.

I'll add testing in a future commit.


bmeneg:
Thinking about it a bit: the default behavior is to show only the MR metadata, I mean, when no flags is passed the command will only show the data specific to the MR, like URL, labels, approves, title, description, etc. Now, when `--comments` is passed both user comments and activities is shown.

Even though it would "break backwards compatibility" I think we should keep `--comments` for what it says: `show comments`, my rational is basically: the word "comments" already implies "user comments", I feel like "activity" is nothing else then an intruder in the `--comments`, understand?

With that, I would go with:
`--comments`: user comments only
`--activities`: system activities only
`--full` or `--all`: user comments + system activities

Also, yesterday I spent some time reading the `git log` documentation and I realized that `--comments=[user,system,all]` (or something like that), defaulting to `all` (to keep the current behavior) isn't _THAT_ bad, even because git does use a lot of this flag style for `--pretty=` and `--decorate=` for instance.

But definitely I would change the current `--comments` behavior :).

@prarit, what do you think?

prarit:
I'm a bit confused :)  @zampierilucas what do you want --comments, --activities to do?  Only show the comments and activities and not the general description, state, etc., of the MR?

zampierilucas:
@prarit I'm partial to bmeneg's idea as it makes clear what will get shown, but I do recognize that this will be as behavior change to the user, as he is used to the `-c` showing all.

So, if also agree, I would like to go with bruno's suggestion.

bmeneg:
@prarit just a clarification: what @zampierilucas is proposing is the behavior described in [1]: differentiate between real user comments vs system notification history (MR state change, new version, labels changing, ...), allowing the user to select one of them or _everything_.

@zampierilucas, I think it's worth adding this link in the commit log.

[1] https://docs.gitlab.com/ee/user/discussions/#show-only-comments

bmeneg:
@zampierilucas also, instead of creating a commit on top of the first commit that brings the same support but in a different version, create just one commit with the final version.

cmd/issue_show.go
```diff
@@ -54,8 +54,17 @@ var issueShowCmd = &cobra.Command{
...
-		showComments, _ := cmd.Flags().GetBool("comments")
-		if showComments {
+		var noteLevel = NoteLevelNone
+
+		if showComments, _ := cmd.Flags().GetBool("comments"); showComments {
+			noteLevel = NoteLevelComments
+		} else if showActivities, _ := cmd.Flags().GetBool("activities"); showActivities {
+			noteLevel = NoteLevelActivities
+		} else if showFull, _ := cmd.Flags().GetBool("full"); showFull || showComments && showActivities {
+			noteLevel = NoteLevelFull
```
bmeneg, line 64: @zampierilucas, the second part of this conditional `|| showComments && showActivities` will never be hit: if any of these were set `noteLevel` would already have been set by previous conditionals.

What I suggest is you check things "backwards": get all the flag values before checking, then check for `showComments && showActivities`, then each of one afterwards.

You got what I mean?

cmd/mr_show.go
```diff
@@ -78,8 +78,17 @@ var mrShowCmd = &cobra.Command{
...
-		showComments, _ := cmd.Flags().GetBool("comments")
-		if showComments {
+		var noteLevel = NoteLevelNone
+
+		if showComments, _ := cmd.Flags().GetBool("comments"); showComments {
+			noteLevel = NoteLevelComments
+		} else if showActivities, _ := cmd.Flags().GetBool("activities"); showActivities {
+			noteLevel = NoteLevelActivities
+		} else if showFull, _ := cmd.Flags().GetBool("full"); showFull || showComments && showActivities {
+			noteLevel = NoteLevelFull
```
bmeneg, line 88: And the same here too.

cmd/issue_show.go
zampierilucas: Yeah, how dumb of me to not notice that :D, I'll send an update

bmeneg (approved):
LGTM

Issue #782, opened 2022-02-03
Unable to create projects in (nested) subgroup


syphernl:
I'm trying to create a bunch of projects in a nested subgroup, which according to #450 should work. 

In my case it however errors out:
```sh
❯ lab project create org/tools/components/foo
2022/02/03 09:58:24 ERROR: project_create.go:57: Group 'org/tools/components' not found
```

The `org/tools/components` subgroup exists.

bmeneg:
Hi @syphernl,

Are you sure your user has write access in the group-level so you can create a new project there?
You must have at least `developer` role and your API token also must have been created with the `api` access checkbox set.

syphernl:
Hi @bmeneg,

Creating projects in the webui works just fine. The token is scoped to `api` and offers `owner` permissions in the Gitlab Group. Should be enough ;-)

I have just tried with a fresh token (which uses the `glpat` prefix) but that one fails as well.

bmeneg:
Ok. That's odd.
Even though I can't reproduce the failure we (I pinged @zampierilucas to help) found something odd in the code indeed.
As soon as I have something I'll open an PR and ask you to test.

bmeneg:
@syphernl would you mind testing the PR #784 changes in your environment ?

syphernl:
> @syphernl would you mind testing the PR #784 changes in your environment ?

@bmeneg I have tested it, it does a bit more than before but I wasn't able to create a project with it just yet.

```
❯ ./lab --version
lab version v0.23.0-56-gb25dc45

❯ ./lab project create org/tools/foo
2022/02/07 13:51:49 ERROR: project_create.go:57: found multiple groups with the same name:
	org/tools/containers
	org/tools/pre-commit-hooks
	org/tools/terraform-aws-components
	org/tools/terraform-modules
	org/tools
use one of the above path options
```

bmeneg:
Hmm, ok. That's indeed broken.
Let me check it again.

bmeneg:
@syphernl ok, can you test it again?

Just a note for the error about `remote origin already exists.`: it happened because you created the project from within a repo dir and lab tried to create an "origin" remote to refer to the newly created project. I added a commit in the PR to remove this _default_ behavior, now, if you really want to create a remote pointing to the new project you can pass the `--remote <name>` flag to the `lab project create` call.

syphernl:
@bmeneg Now I'm getting an error that the group doesn't exist:

```sh
❯ ./lab/lab --version
[...]

PR #784, opened 2022-02-03
project_create: fix group lookup and allow custom remote name

bmeneg:
The group search logic didn't consider the fact that the GitLab API
returns an alphabetical ordered list of groups within a certain group,
thus when multiple subgroups were nested the value groups[0] and list[0]
used in the code weren't always the expected ones.

With that, instead of looking up for the first group, we look for the
last one, meaning that all subgroups found in user's namespace or
wherever he has access to, is returned from the API. The match then
happens with the group full path or full name, which matches first.

In case the user has access to multiple groups with ambiguous names,
all possible paths are shown and lab's exits.

Also, a new flag `--remote` was added to allow the user to provide a
remote name in case needed. `lab` don't add a remote by default.

Related: #782

### DIFF
--- a/cmd/project_create.go
+++ b/cmd/project_create.go
@@ -24,6 +24,7 @@ var projectCreateCmd = &cobra.Command{
 	Example: heredoc.Doc(`
 		lab project create myproject
 		lab project create myproject -n "new proj"
+		lab project create myproject -r myupstream
 		lab project create -g mygroup myproject
 		lab project create mygroup/myproject -n "new proj"
 		lab project create myproject --http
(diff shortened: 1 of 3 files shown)


cmd/project_create.go
```diff
@@ -123,6 +131,7 @@ func init() {
 	projectCreateCmd.Flags().StringP("name", "n", "", "name of the new project")
 	projectCreateCmd.Flags().StringP("group", "g", "", "group name (also known as namespace)")
 	projectCreateCmd.Flags().StringP("description", "d", "", "description of the new project")
+	projectCreateCmd.Flags().StringP("remote", "r", "", "add remote referring to the new project")
```
zampierilucas, line 134: I think we could use one example of that

bmeneg: Done! :)

Issue #787, opened 2022-02-06
host not being set if caFile is set.


jcengland:
After setting the caFile directive in the `lab.toml` file due to using a self signed cert I get the following error:
```
15:40 $ lab project list
2022/02/04 15:40:25 ERROR: project_list.go:57: Get "/api/v4/projects?membership=false&order_by=id&owned=false&per_page=100&search=&simple=true&sort=asc&starred=false": unsupported protocol scheme ""
```
`host` is set in `lab.toml` :
```

[core]
  host = "https://example.com/gitlab/"
  token = "XXXXXXXXXXXXXXX"

  user = "england"

[tls]
  ca_file = "/Users/england/.config/lab/cert.pem"
```
But this did look like `host` was not set.  Let me say I know little about go so this could be way off base so take this with a grain of salt.  It seems that in the function `InitWithCustomCA` has `_host` passed in but `host = _host` is never set as it is in the `Init` function.  Looks like this code has been there for a while so I might be missing something.  

Thanks for working on this project, it is really useful.

Clay

edit: this was with the current release v0.23.0

bmeneg:
Hey @jcengland , thanks for the report!

And yes. You're right.
I think it was never really spotted before just because no one really used it before neither, obviously, tested (what surprises me tbh).

I'm going to submit a PR and let you test it before merging.

bmeneg:
@jcengland would you mind testing #788 in your environment?

jcengland:
That worked!  Thanks for getting that fixed.

Clay

PR #788, opened 2022-02-07
internal/gitlab: fix init with custom CA file

bmeneg:
None of the internal/API user's information (host, user, token) were
being set when a custom CA file was set in the cofiguration file. This
patch set them the same way as the Init() function.

Related: #787


bmeneg:
Since the change is quite simple and the user already tested, I'm merging it.

Issue #381, opened 2020-06-29
Git internals don't work remote oauth url


mrbrownt:
I am using an oauth remote url in the format `https://oauth2:${SERVICE_ACCOUNT_GITLAB_TOKEN}@gitlab.com/${CI_PROJECT_PATH}.git`

When running `lab mr create ${BRANCH} master` I'm getting the following error.
```
mr_create.go:83: the key `remote.https://oauth2:[MASKED]@gitlab.com/some/repo.git.url` is not found
```
I know this upstream remote url does exist because it's was used to create a branch prior to this command.
I'm running this all in gitlab CI.

zaquestion:
@mrbrownt Can you share the git config for this and verify what version of lab you're running? It's very weird that we'd return the error like that. The lookup in play is trying to determine the project path from the remote url.

Is `${BRANCH}` in this case referring to: `https://oauth2:${SERVICE_ACCOUNT_GITLAB_TOKEN}@gitlab.com/${CI_PROJECT_PATH}.git` ?

Just to be clear, the create usage is as below, and refers to the name of the remote in your git config.
```
create [remote [branch]]
```

That said you don't seem to be making it far enough for the arguments you're providing to actually come into play. Based on the error output, my guess is there is some git config that looks like 
```
[branch "mybranch"]
	remote = https://oauth2:${SERVICE_ACCOUNT_GITLAB_TOKEN}@gitlab.com/${CI_PROJECT_PATH}.git
	merge = refs/heads/mybranch
```
As this logic: https://github.com/zaquestion/lab/blob/a7646da14338df8cdaabc2a9f80c45d47c2bd322/cmd/mr_create.go#L179-L183 is the only way I could see it possible for https://github.com/zaquestion/lab/blob/a7646da14338df8cdaabc2a9f80c45d47c2bd322/cmd/mr_create.go#L81-L84 to fail with the error you're seeing.

I would have thought the above git config would be invalid, so if that's not the case and `remote` can be equal to a url, then that would be the bug we need to address in lab and we'd need to be a bit surgical to support it everywhere it needs to be. A workaround would probably be to add the remote definition in a way lab expects

```
[remote "oauthurl"]
	url = https://oauth2:${SERVICE_ACCOUNT_GITLAB_TOKEN}@gitlab.com/${CI_PROJECT_PATH}.git
	fetch = +refs/heads/*:refs/remotes/oauthurl/*
[branch "mybranch"]
	remote = oauthurl
[...]

PR #793, opened 2022-02-11
remote: support remotes referred by URL only

bmeneg:
When using git-push the user can set the upstream of a branch to a
remote that is not present in the current repository by referring to its
URL. In these cases, git checks for the validity of the URL itself
instead of checking if a remote with that name exists locally.

I.e.:
```
$ git push --set-upstream <URL>
```

This case affects lab mostly when we use the remote URL to lookup for it
in GitLab: lab checks for the branch's remote name.
`branch.<branch>.remote`, then it gets the remote URL itself; however,
in this situation, the `branch.<branch>.remote` will be the URL already
and any check for that remote in the local repository fails.

The biggest problem is the amount of different URL formats supported by
Git [1], which is bigger than the ones supported by the usual
url.Parse() (net/url Go module). Because of that, I brought the git-urls
module to help.

This new module threat anything that isn't an external URL as a possible
file URL, meaning that default remote names are also accepted, with that
we can pass all our remote names (URL or not) to its Parse() function
and just use the "path" component, which is pretty close to what we need
for the GitLab lookup.

[1] https://git-scm.com/docs/git-push#URLS

Related: #381

### DIFF
--- a/cmd/ci_run.go
+++ b/cmd/ci_run.go
@@ -100,7 +100,10 @@ func getCIRunOptions(cmd *cobra.Command, args []string) (string, string, error) 
 
 	var pid string
 
-	remote := determineSourceRemote(branch)
+	remote, err := determineSourceRemote(branch)
+	if err != nil {
+		return "", "", err
+	}
 	rn, err := git.PathWithNamespace(remote)
 	if err != nil {
 		return "", "", err
--- a/cmd/mr_create.go
+++ b/cmd/mr_create.go
@@ -139,7 +139,11 @@ func runMRCreate(cmd *cobra.Command, args []string) {
 		log.Fatal(err)
 	}
 
-	sourceRemote := determineSourceRemote(localBranch)
+	sourceRemote, err := determineSourceRemote(localBranch)
+	if err != nil {
+		log.Fatal(err)
+	}
+
 	// Get the pushed branch name
 	sourceBranch, _ := git.UpstreamBranch(localBranch)
 	if sourceBranch == "" {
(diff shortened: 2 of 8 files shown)


bmeneg:
I'll move this MR to draft until I get the tests fixed.

prarit:
@bmeneg , it's always easier for me to see a real world example.  Can you show me how this would actually work within a git repo?

bmeneg:
@prarit yes sure.

One practical usage of it is on CI environments where remote names are not really required/used.

But something simpler: the user can setup a repository and for whatever reason push things so some repository using the remote URL instead of the same itself. For instance:
```
$ git clone https://github.com/bmeneg/lab
$ git checkout fix-awesome-feature
<.. do some work ..>
$ git push https://github.com/zaquestion/lab
```
works just fine.

At the same time, the user can set that remote URL as a real tracking remote:
```
$ git push -u https://github.com/zaquestion/lab test_branch
```
Causing the `branch.test_branch.remote` to be an URL instead of an usual name, as we are more used to do, thus, every subsequent `git push` would push to that URL directly, meaning no remote name lookup nor anything, just straight push to a branch in that repo URL.

I personally don't use remote URLs directly, but Git does support it and mention it in the first lines of the git-push help:
```
<repository>

    The "remote" repository that is destination of a push operation. This parameter can be either a URL
    (see the section [GIT URLS](https://git-scm.com/docs/git-push#URLS) below) or the name of a remote 
    (see the section [REMOTES](https://git-scm.com/docs/git-push#REMOTES) below).
```

bmeneg:
Ah, one thing that might help a bit to understand is: we need to distinguish between a remote name vs url because today we use it to retrieve GitLab's project:

```
remote_name = branch.<branch_name>.remote
remote_url = remote.<remote_name>.url
gitlab.GetProject(remote_url)
```
however, in case we don't have a real remote name, we have to skip the second step from above, ending up with the following flow:

```
branch.<branch_name>.remote = remote_url
gitlab.GetProject(remote_url)
```

zampierilucas (approved):

Issue #789, opened 2022-02-10
shell completion command failed when cwd is local git repo


georgexsh:
I am using lab version 0.23.0, bash 5.1, steps to reproduce this issue:

- create a git repo without remote: `RAND=$RANDOM; mkdir /tmp/$RAND; cd /tmp/$RAND; git init`
- run `source <(lab completion)`
- got error: `2022/02/10: No such file or directory`

`lab completion` [outputs a line of log](https://github.com/zaquestion/lab/blob/8ab707a74bee416ba35090208cc4c7f6f67c8c2d/cmd/root.go#L203) to `stdout` like `2022/02/10 13:07:32 root.go:205: No default remote found`, which bash's `source` doesn't understand. I guess a possible solution is [always output log to `stderr` rather than `stdout`](https://github.com/zaquestion/lab/blob/8ab707a74bee416ba35090208cc4c7f6f67c8c2d/internal/logger/logger.go#L49), or bypass local git remote checking when the command is `lab completion`.

bmeneg:
@georgexsh #792 should solve the issue with `lab completion`.

bmeneg:
I'm going to close this issue.
But in case you see any other problems, please feel free to reopen it or open a new issue!

Many thanks for the report @georgexsh

georgexsh:
@bmeneg the fix works, thanks! but I still wonder why the logs would be writing to `stdout`.

bmeneg:
@georgexsh That's a valid point: not sure why, but when I wrote the logs I let only the "error" level to be written to stderr, while others "info, warn and debug" would go stdout.

But yes, anything that isn't part of the program's standard output, shouldn't go to stdout.
I'm going to prepare another PR suggesting this change.

PR #794, opened 2022-02-12
internal/logger: send any logging message to stderr

bmeneg:
Our internal logger was sending Error messages to stderr and all the
other levels (Info, Warn and Debug) to stdout. However, both POSIX stds
and part of the utility development community that "diagnostic" messages
should go only to stderr, leaving stdout for the actual results of the
program. Even though the logger messages are quite differet (prefixed by
date, file and line) from lab's actual results, it forces a program
receiving the stdout though a, eg., pipe to filter logger messages if
only the actual results are required.

One example of this issue was presented in lab's issue 789 [1], which
was solved in a different way, but failed because a "diagnostic" message
was feed into the `source` utility.

This patch redirects all logging messages (not actual lab results) to
stderr and also remove the function SetStdDest() that wasn't being used
anywhere in the code.

[1] https://github.com/zaquestion/lab/issues/789


prarit (approved):
@bmeneg this makes sense to me and looks like something we should have changed a while ago :)

PR #795, opened 2022-02-15
lab: Provide a better "Project Not Found" error message

prarit:
Currently lab outputs this error when an unknown project is found:

"GitLab project not found, verify you have access to the requested
resource".

While this error does match the webUI, it doesn't provide enough
information to debug the error.  The webUI in this case still provides
the name of the unknown project where lab only outputs the generic
warning.

Update the "Project Not Found" error with the unknown project name.


prarit:
I have a hunch this will fail testing.

PR #802, opened 2022-02-18
lab: Add MR and Issue note delete functionality

prarit:
Add the ability to delete notes from issues and MRs with the --deletenote
option.

### DIFF
--- a/cmd/issue_edit.go
+++ b/cmd/issue_edit.go
@@ -26,7 +26,8 @@ var issueEditCmd = &cobra.Command{
 		lab issue edit 14 -m "new title" -m "new desc"
 		lab issue edit 14 -l new_label --unlabel old_label
 		lab issue edit --milestone "NewYear"
-		lab issue edit --force-linebreak`),
+		lab issue edit --force-linebreak
+		lab issue edit --delete-note 14:2065489`),
 	Args:             cobra.MinimumNArgs(1),
 	PersistentPreRun: labPersistentPreRun,
 	Run: func(cmd *cobra.Command, args []string) {
@@ -52,6 +53,35 @@ var issueEditCmd = &cobra.Command{
 		issue, err := lab.IssueGet(rn, issueNum)
 		if err != nil {
 			log.Fatal(err)
+		}
+
+		deleteNote, err := cmd.Flags().GetBool("delete-note")
+		if err != nil {
+			log.Fatal(err)
+		}
+		if deleteNote {
+			discussions, err := lab.IssueListDiscussions(rn, int(issueNum))
+			if err != nil {
+				log.Fatal(err)
+			}
+
+			discussion := ""
+		findDiscussionID:
+			for _, d := range discussions {
+				for _, n := range d.Notes {
+					if n.ID == commentNum {
+						discussion = d.ID
+						break findDiscussionID
+					}
+				}
+			}
+
+			// delete the note
+			err = lab.IssueDeleteNote(rn, issueNum, discussion, commentNum)
+			if err != nil {
+				log.Fatal(err)
+			}
+			return
 		}
 
 		linebreak, err := cmd.Flags().GetBool("force-linebreak")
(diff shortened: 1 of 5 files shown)


zampierilucas:
@prarit Can you add some usage examples to the doc?

prarit:
Yep, fixed and pushed.

zampierilucas (approved):

cmd/issue_edit.go
```diff
@@ -26,7 +26,8 @@ var issueEditCmd = &cobra.Command{
 		lab issue edit 14 -m "new title" -m "new desc"
 		lab issue edit 14 -l new_label --unlabel old_label
 		lab issue edit --milestone "NewYear"
-		lab issue edit --force-linebreak`),
+		lab issue edit --force-linebreak
+		lab issue edit 14 --deletenote <noteID>`),
```
bmeneg, line 30: I have the impression that this help message is completely wrong 😄 .
It should be `lab issue edit [remote] <mr_id>:<comment_id>`. the example leads to an `invalid remote` sort of error.
And the same happens in the mr_edit.go code.

cmd/issue_edit.go
```diff
@@ -54,6 +55,34 @@ var issueEditCmd = &cobra.Command{
...
+			if err != nil {
+				log.Fatal(err)
+			}
+
+			discussion := ""
+			for _, d := range discussions {
+				for _, n := range d.Notes {
+					if n.ID == commentNum {
+						discussion = d.ID
+						break
```
bmeneg, line 73: This break will break the inner loop only, but the outer will keep running and, consequently, run the inner for the new discussions. You can either use a loop label or check if `discussion != ""` in the end of the outer loop.

Label example:
```go
myLabel:
    for _, d := range discussions {
        for _, n := range d.Notes {
            if n.ID == cummentNum {
                discussion = d.ID
                break myLabel
            }
        }
    }
```

The same happens in the mr_edit.go code.

bmeneg (changes requested):

cmd/issue_edit.go
```diff
@@ -54,6 +55,34 @@ var issueEditCmd = &cobra.Command{
 			log.Fatal(err)
 		}
 
+		deleteNote, err := cmd.Flags().GetBool("deletenote")
```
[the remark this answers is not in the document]
zampierilucas: I had the same impression while reviewing, but at the end of it I wished we could just use `--delete` but I'm afraid it might be ambiguous to other notes fields(instead of the note itself).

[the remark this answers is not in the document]
prarit: Oh good idea.  I've changed this to 'delete-note'.

cmd/mr_edit.go
```diff
@@ -33,7 +33,8 @@ var mrEditCmd = &cobra.Command{
 		lab mr edit 31 upstream --draft
 		lab mr edit 37 upstream --ready
 		lab mr edit 41 upstream -r johndoe -r janedoe
-		lab mr edit 43 upstream --unreview johndoe`),
+		lab mr edit 43 upstream --unreview johndoe`
```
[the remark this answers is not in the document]
prarit: Whoops. Fixed.

cmd/issue_edit.go
prarit: Oh of course.  That has to be 'lab issue edit <id>:<comment_id>'.  I've fixed it in both cases.

cmd/issue_edit.go
prarit: Fixed (and new version pushed).

cmd/issue_edit.go
```diff
@@ -26,7 +26,8 @@ var issueEditCmd = &cobra.Command{
 		lab issue edit 14 -m "new title" -m "new desc"
 		lab issue edit 14 -l new_label --unlabel old_label
 		lab issue edit --milestone "NewYear"
-		lab issue edit --force-linebreak`),
+		lab issue edit --force-linebreak
+		lab issue edit 14 --delete-note <issueID>:<noteID>`),
```
bmeneg, line 30: The `14` already is the `<issueID>` isn't it? Just remove it. The same is valid for the mr_edit.go message.

What about using "real" values instead of the placeholders? 
For instance: `lab issue edit --delete-note 14:736152913`

bmeneg (changes requested):

cmd/issue_edit.go
prarit: Oh duh.  Yes.  Fixed.

bmeneg (approved):

bmeneg:
@prarit now the build test broke due to gofmt on the label thing.
But other than that. LGTM.

prarit:
> @prarit now the build test broke due to gofmt on the label thing. But other than that. LGTM.

Grr..  fixed.

PR #812, opened 2022-02-22
cmd/issue: add "i" alias to the command

### LINKED ISSUE #810: `issue`: add alias for just `i`?
I hesitate to make this request b/c it might be workflow specific to me (or just me being laaaazy), but I use `lab issue` a lot and would love to have a built-in alias for just `lab i`. 

**Considerations**  
* aliases are currently used only for subcommands (ie `lab issue b` for `lab issue browse`) not top level commands
  * exceptions includes `lab repo` (alias for `lab project`) and `lab snip` (alias for `lab snippet`)
  * So it's a little bit unprecedented to add a alias of this kind to a top level command.
* Right now, when I run `lab i`, it gets handed off to `git` and the only `i` based command that git suggests is `init`. 
  * So there aren't any/many `git i*` or `lab i*` commands that could confuse this.

**Workaround**
I currently have a git alias defined for this: `i = "!lab issue"`, and this works, but it triggers the "lab's git passthrough command support will be removed in a later release." warning whenever I use it. 

Is there a way to create custom command aliases just w/i lab, or is the only way to do this w/ a code change w/i `lab`?

bmeneg:
The 'issue' command is one of the high-demand commands within lab with,
probably, 'mr', and enabling the alias 'i' for it seems a good approach
for all lazy developers that we all are. 'mr', 'todo', 'ci', 'repo' are
already shorthand themselves, adding this specific to 'issue' should not
harm, even more considering we aren't creating one-letter aliases to
every single command.

Fix #810 .

### DIFF
--- a/cmd/issue.go
+++ b/cmd/issue.go
@@ -6,6 +6,7 @@ import (
 
 var issueCmd = &cobra.Command{
 	Use:              "issue",
+	Aliases:          []string{"i"},
 	Short:            `Describe, list, and create issues`,
 	Long:             ``,
 	PersistentPreRun: labPersistentPreRun,


zampierilucas:
I'm foreshadowing people wanting this to be configurable 😬

zampierilucas (approved):

bmeneg:
> I'm foreshadowing people wanting this to be configurable 😬

Easy answer: alias it locally.

Hotkeys I understand people willing to configure, but command shorthand is too much. LOL.

prarit:
"i".  LOL

prarit (approved):

PR #821, opened 2022-04-06
cmd/mr_create: add --no-edit flag to not open editor

### LINKED ISSUE #819: mr create: accept default MR description
We open new MRs with a command like:
```
$ lab mr create --force-linebreak --draft <remote> <target_branch>
```
and always use the default description that is loaded in the editor.  Is there a way to accept the default MR description from the command line?

Alternatively, should we just do something like?
```
$ lab mr create -F <(git show -1) ...
```

bmeneg:
The default behavior when creating an MR is to open the editor regardless
the amount of commits:

- 1 commit: open editor with the commit's title and body already in place.
- 1+ commits: open editor with an empty message

This patch adds the --no-edit to allow users of the first option (single
commit) to tell lab to create the MR using the commit's message right away,
without opening the editor.

If --no-edit is used with the second option (multiple commits) lab exits
with a error message.

Fixes #819 

### DIFF
--- a/cmd/mr_create.go
+++ b/cmd/mr_create.go
@@ -57,6 +57,7 @@ func init() {
 	mrCreateCmd.Flags().String("milestone", "", "set milestone by milestone title or ID")
 	mrCreateCmd.Flags().StringP("file", "F", "", "use the given file as the Title and Description")
 	mrCreateCmd.Flags().StringP("file-edit", "f", "", "use the given file as the Title and Description and open the editor")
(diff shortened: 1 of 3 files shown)


prarit (approved):
Looks good to me.  My apologies for the delay in review.

zampierilucas (approved):

bmeneg:
The test failure seems related to the PR https://github.com/zaquestion/lab/pull/826

PR #825, opened 2022-04-09
mr list: Milestone exact match fix

prarit:
The following error occurs with 'lab mr list' when looking for MRs with a
specific milestone:

[prarit@prarit rhel-8]$ lab mr list --state merged --milestone 'RHEL-8.7.0'
2022/04/09 08:58:45 ERROR: mr_list.go:120: Milestone 'RHEL-8.7.0' is ambiguous

There are two milestones with similar names in the project:

[prarit@prarit rhel-8]$ lab milestone list  | grep "8.7.0"
RHEL-8.7.0-test - Red Hat Enterprise Linux 8, release 8.7.0
RHEL-8.7.0 - Red Hat Enterprise Linux 8.7.0 release

The lab error occurs because the GitLab API List Milestones option does
not do exact matching and returns a list of possible matching milestones.
Since no one else has hit this error it seems safe to look for and return
an exact match if the API returns multiple milestones.

Look for an exact match when searching for a milestone.


zampierilucas:
I'm not sure if I agree with this one, as I can see myself using vague terms as --milestone "8.x" and wanting to get the result for that stream without having to specify the prefix.

That being said, It seems like we are using `GET /projects/:id/milestones?search=<lab input>`, on line 913, to get milestones back, I propose we use `GET /projects/:id/milestones?title=<lab input>` when the user wants to be specific with his request.

I'm not sure if another option would suit it, perhaps --title ?

zampierilucas:
Instead of checking it agains the user input, you could change line 913 to title instead of serach, that would make gitlab ensures exact match instead of us.

bmeneg:
> Instead of checking it agains the user input, you could change line 913 to `title` instead of `search`, that would make GitLab ensures exact match instead of us.

@prarit I agree with Lucas. GitLab already offers the exact match logic with "Title" instead of "Search". So changing it in `&gitlab.ListMilestonesOptions` and removing the "ambiguity" conditional seems better.

prarit:
Fixed.  Sorry for the delay.

zampierilucas (approved):

bmeneg (approved):

PR #837, opened 2022-06-23
Fix guessing the MR ID based on the current branch

bmeneg:
This PR fixes the lab's "guess" operation for retrieving the MR ID
based on the currently checked-out branch, which was, recently,
causing issues with the testing code.

Currently, if any MR opened from a forked project but with the 
same source branch name will get mixed with MRs from the 
original project with that name, possibly returning the wrong
ID to the user.

Check the second commit for more detailed information.

### DIFF
--- a/internal/gitlab/gitlab.go
+++ b/internal/gitlab/gitlab.go
@@ -453,29 +453,19 @@ func MRGet(projID interface{}, id int) (*gitlab.MergeRequest, error) {
 // MRList lists the MRs on a GitLab project
 func MRList(projID interface{}, opts gitlab.ListProjectMergeRequestsOptions, n int) ([]*gitlab.MergeRequest, error) {
 	if n == -1 {
-		opts.PerPage = maxItemsPerPage
+		n = maxItemsPerPage
 	}
 
-	list, resp, err := lab.MergeRequests.ListProjectMergeRequests(projID, &opts)
-	if err != nil {
-		return nil, err
-	}
-
-	var ok bool
-	if opts.Page, ok = hasNextPage(resp); !ok {
-		return list, nil
-	}
-
-	for len(list) < n || n == -1 {
-		if n != -1 {
-			opts.PerPage = n - len(list)
-		}
+	var list []*gitlab.MergeRequest
+	for len(list) < n {
+		opts.PerPage = n - len(list)
 		mrs, resp, err := lab.MergeRequests.ListProjectMergeRequests(projID, &opts)
 		if err != nil {
 			return nil, err
 		}
 		list = append(list, mrs...)
 
+		var ok bool
 		if opts.Page, ok = hasNextPage(resp); !ok {
 			break
 		}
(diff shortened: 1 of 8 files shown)


zampierilucas (approved):

bmeneg:
Unfortunately, I'm not being able to figure out what's going on with the test: why it's failing. Give me a few more hours and I'll have it fixed.

prarit (approved):

bmeneg:
Ok, I had to add an additional commit to fix a few test cases that were changed due to old bogus tests that left the test repo in a weird state (ie. MRs were opened but not closed)

internal/gitlab/gitlab.go
```diff
@@ -463,6 +463,18 @@ func MRList(projID interface{}, opts gitlab.ListProjectMergeRequestsOptions, n i
 		if err != nil {
 			return nil, err
 		}
+		// SourceBranch was specified, so we need to make sure we're listing MRs
+		// for the source branch for the specific projID and not a fork.
```
fmuellner, line 467: That assumes that the lab user only creates merge requests from branches on the upstream remote, and not their own fork. If you are pushing to a fork and create a merge request from that, then the whole feature is now broken.

IMHO the correct ID for filtering is the project that corresponds to the branch's upstream remote.

Maybe including `AuthorID` in the options would be an alternative to filtering by branch, although that would mean that guessing doesn't work for other people's merge requests (for example for branches created via `lab mr checkout ...`).

fmuellner: > Do you mean the following change?

No. As I understand it, `SourceProjectID` refers to the project where the branch that is being merged comes from, and `ProjectID` is the project where it is being merged (no clue what's the difference between `ProjectID` and `TargetProjectID`).

If that's correct, then it depends on whether the two should or should not match:

.1 Push directly to the upstream remote, IDs should match:
```sh
$ git push -u origin current-branch; lab mr create
```
2. Push to the personal fork, IDs should not match:
```sh
$ git push -u fmuellner current-branch; lab mr create
```

I think what we *really* want is based on the branch's upstream remote (like `git config branch.current-branch.remote`) and somehow match that to a gitlab project. I'm afraid I don't have any great ideas how to actually do that :-(

In particular "origin" is tricky, as I've seen two different uses:
 - 'origin' refers to the upstream
 - 'origin' refers to the personal fork, the upstream remote uses 'upstream'

We probably have to look at the remote's url. It looks like the Project type has `SSHURLToRepo` and `HTTPURLToRepo` fields, but relying on exact matches sounds fragile (I'm thinking of stuff like http vs. https, trailing slashes, ...)

fmuellner: > I think what we _really_ want is based on the branch's upstream remote (like `git config branch.current-branch.remote`) and somehow match that to a gitlab project. I'm afraid I don't have any great ideas how to actually do that :-(

It turns out that all the bits that are needed are already in place, so this was a lot easier than I thought:
https://github.com/zaquestion/lab/pull/838

bmeneg: Ah right! I see what you meant and how it breaks `mr create`.
Thanks for the quick PR. I'll review #838.

PR #838, opened 2022-07-01
Consider branch remote when guessing MR

fmuellner:
When trying to guess the merge request that corresponds to a
local branch, we search for MRs that match the (upstream) branch
name.

But as branch names only have to be unique within a project, it
is possible that there are multiple merge requests from different
forks that use the same source branch name.

Luckily we do know which is the correct source project: It's the
one that corresponds to the remote that the local branch was pushed
to.

### DIFF
--- a/cmd/mr_browse_test.go
+++ b/cmd/mr_browse_test.go
@@ -20,7 +20,7 @@ func Test_mrBrowseWithParameter(t *testing.T) {
 }
 
 func Test_mrBrowseCurrent(t *testing.T) {
-	git := exec.Command("git", "checkout", "mrtest")
+	git := exec.Command("git", "checkout", "mrtest2")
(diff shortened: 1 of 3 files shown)


bmeneg:
This PR changes the expected behavior by the Test_mrBrowseCurrent test code: the `mrtest` branch tracks the `lab-testing` remote, while the test expects an MR (!1) which has as its source project the upstream own `mrtest` branch.

To fix it I pushed to your branch the necessary changes in  `cmd/mr_browse_test.go`. Hopefully, it'll be enough.

bmeneg (approved):

bmeneg:
I'll also create a "hotfix" with your change and release it as 0.25.1.
Many thanks, @fmuellner!

Issue #206, opened 2018-08-10
Support project group operations with `-g` on `lab clone` and `lab project list`


zaquestion:
Following on from #185 we want to support listing and cloning groups directly without having to parse `lab project list` results

krlmlr:
Also for `lab project create` (#211)?

dosmanak:
Hello. I put this tiny change into mentioned PR. It is only for `project list`.

dosmanak:
Hello. It seems there is not much activity in this project :crying_cat_face:

zampierilucas:
@dosmanak most development has moved to https://github.com/aristocratos/btop, as it is the next iteration of bpytop

alerque:
What does `btop` have to do with `lab`?

zampierilucas:
@alerque wow, it seems I really messed up some tabs at some point, one was nothing to do with the other really,

alex0ptr:
Are you still interested in supporting this? I think we could provide a PR.

dosmanak:
@alex0ptr I am not sure what do you mean. I've prepared the PR #842 to list projects within groups. It is tiny but working.
for `lab clone` a PR would be nice.

alex0ptr:
Sorry I meant `lab clone` of course. Ok, I'll look into it in the coming weeks.

PR #842, opened 2022-07-19
feat: #206 add flag to list projects also matching group name

dosmanak:


dosmanak:
The test failes becaut the test repo is different than expected. That is not my fault ;-)

prarit:
@dosmanak , sorry for the delay in response.  I'm not sure all your code is in this MR?  AFAICT you've only added the -g option but I don't see what you do with it.

cmd/project_list.go
```diff
@@ -44,13 +46,14 @@ var projectListCmd = &cobra.Command{
...
-			Membership: gitlab.Bool(projectListConfig.Membership),
-			Starred:    gitlab.Bool(projectListConfig.Starred),
-			Search:     gitlab.String(search),
+			Simple:           gitlab.Bool(true),
+			OrderBy:          gitlab.String("id"),
+			Sort:             gitlab.String("asc"),
+			Owned:            gitlab.Bool(projectListConfig.Owned),
+			Membership:       gitlab.Bool(projectListConfig.Membership),
+			Starred:          gitlab.Bool(projectListConfig.Starred),
+			SearchNamespaces: gitlab.Bool(projectListConfig.Group),
```
dosmanak, line 55: I simply pass the flag `-g` as boolean parameter wich sets the `SearchNamespaces` boolean for the gitlab SDK github.com/xanzy/go-gitlab [SDK](https://pkg.go.dev/github.com/xanzy/go-gitlab#ListProjectsOptions)

which translates into gitlab [API](https://docs.gitlab.com/ee/api/projects.html#list-all-projects) call.

dosmanak: Is it clear?

PR #846, opened 2023-01-26
Don't load config during shell completion

ralismark:
Tab completing `lab` causes it to try and load the config, which fails if it hasn't been initialised yet, causing it to produce an error and fail to show any completions. This is because the completion invokes `lab _carapace`, which wasn't correctly detected by `skipInit`.

### DIFF
--- a/main.go
+++ b/main.go
@@ -28,6 +28,10 @@ func main() {
 }
 
 func skipInit() bool {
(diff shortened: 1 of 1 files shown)


ralismark:
should this be backported?

prarit:
Thanks for the contribution! :)

PR #856, opened 2023-05-22
fix(browse): Update browse URLs

claytonrcarter:
Like #855, this also deals with a breaking change in GL 16.0. In this case, support for "legacy" URLs has been removed, and this is also breaking `lab issue browse`, `... mr browse`, and `... snippet browse`. This change should update all of these so that they work again.

**BREAKING**
As I understand it, these delimited URLs should work on any version of GL since v9 (2017) but these changes will bread the browse commands for users of older versions. If this is a concern, I suppose we could use a config flag to control the URL format.

**FMI** 
Deprecation notice: https://docs.gitlab.com/ee/update/deprecations.html?removal_milestone=16.0#legacy-urls-replaced-or-removed
List of affected URLs: https://gitlab.com/gitlab-org/gitlab/-/issues/28848#release-notes

### DIFF
--- a/cmd/issue_browse.go
+++ b/cmd/issue_browse.go
@@ -31,7 +31,7 @@ var issueBrowseCmd = &cobra.Command{
 
 		// path.Join will remove 1 "/" from "http://" as it's consider that's
 		// file system path. So we better use normal string concat
-		issueURL := project.WebURL + "/issues"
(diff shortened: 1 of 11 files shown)


zampierilucas (approved):

claytonrcarter:
This has been rebased onto #857 and is pointing at that branch. This should be merged after that one.

PR #858, opened 2023-05-23
go modules update 20230523

prarit:
Update go modules with 'go get -u ./...' and fix the fallout from the module updates.


zampierilucas (approved):

zampierilucas (approved):

claytonrcarter:
This is great! I have a local build that updates `go-gitlab` to 0.74 (I think ... I did that to support some local customizations that add support for `--assignee=none` and a few others), but I never got around to sharing. It looks like you've already got it, though!

PR #859, opened 2023-05-24
internal/config/config.go: Warn on expiring tokens

prarit:
82444bfa9c07 ("config: Add config verification") added some simple checking to see if a user's token was valid.  After the recent go module updates we can now use the PersonalAccessToken service from go-gitlab to warn a user that their token is about to expire.  This warning should happen once a day when the config is verified.

Warn a user when their token is going to expire within two weeks.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -191,10 +191,10 @@ func checkTokenAndGetUser(host, token string, skipVerify bool) string {
 	loc, _ := time.LoadLocation("UTC")
 	checkTime := time.Now().UTC()
 	midnightUTC := MainConfig.GetTime("core.TokenCheckTime")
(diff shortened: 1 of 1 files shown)


claytonrcarter (approved):
LGTM ... I think this will be really helpful!

claytonrcarter:
BTW the build error looks like the commit just needs `go fmt`

claytonrcarter:
FYI I think I borked the tests while working on the tests for #860 ... I updated a few of the issues and MRs in the test project and I'm still untangling a couple of the test failures. There's a wip commit in #860 w/ the updates needed, but it's not done yet. Much apologies!

claytonrcarter:
I suspect that it is, yes. And if not, a fix is just a commit away. 😂

PR #869, opened 2023-07-13
cmd/project_browse.go: Add --file and --branch options

### LINKED ISSUE #866: support browsing to project file
`lab project browse` currently opens the project root in the browser. It would be useful to allow for a means to open a specified project file in the browser, e.g. `lab project browse --file ./path/to/file.go`

prarit:
@sding3 has requested the ability to browse to a specific file using a --file argument.  In addition to this I think it is a good idea to also add a --branch argument so users can browse a file on a specific branch.

Add --file and --branch options.

Close #866

### DIFF
--- a/cmd/project_browse.go
+++ b/cmd/project_browse.go
@@ -1,15 +1,44 @@
 package cmd
 
 import (
+	"github.com/MakeNowJust/heredoc/v2"
 	"github.com/spf13/cobra"
 	lab "github.com/zaquestion/lab/internal/gitlab"
 )
 
+var (
+	projectFile    string
+	projectBranch  string
+	projectFileRef string
+)
+
+func projectBrowseGetPath(webURL string, defaultBranch string, branch string, file string, ref string) (path string) {
+	if branch == "" {
+		branch = defaultBranch
+	}
+
+	if ref != "" {
+		path = webURL + "/-/tree/" + ref
+		return path
+	}
+
+	if file != "" {
+		path = webURL + "/-/blob/" + branch + "/" + file
+	} else {
+		path = webURL + "/-/tree/" + branch
+	}
+
+	return path
+}
+
 var projectBrowseCmd = &cobra.Command{
-	Use:              "browse [remote]",
-	Aliases:          []string{"b"},
-	Short:            "View project in a browser",
-	Example:          "lab project browse origin",
+	Use:     "browse [remote]",
+	Aliases: []string{"b"},
+	Short:   "View project in a browser",
+	Example: heredoc.Doc(`
+		lab project browse origin
+		lab project browse --file arch/x86/Makefile
+		lab project b --ref ce697ccee1`),
 	PersistentPreRun: labPersistentPreRun,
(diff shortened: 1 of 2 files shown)


cmd/project_browse.go
```diff
@@ -22,13 +44,15 @@ var projectBrowseCmd = &cobra.Command{
...
+		err = browse(projectBrowseGetPath(p.WebURL, p.DefaultBranch, projectBranch, projectFile))
 		if err != nil {
 			log.Fatal(err)
 		}
 	},
 }
 
 func init() {
+	projectBrowseCmd.Flags().StringVar(&projectFile, "file", "", "path to specified file")
+	projectBrowseCmd.Flags().StringVar(&projectBranch, "branch", "", "specific branch (overrides default branch)")
```
sding3, line 56: Rather than a branch, could this take a git ref such that it would work if the user supplied a git tag or a git SHA for example?

Also, when omitted to, could this default to `git rev-parse HEAD` of the git repo that the calling process is currently located?

prarit: >Rather than a branch, could this take a git ref such that it would work if the user supplied a git tag or a git SHA for example? 

Sure, can you provide me an example URL so I can reference it?

> Also, when omitted to, could this default to git rev-parse HEAD of the git repo that the calling process is currently located?

This gets very tricky when dealing with local git repositories.  'git rev-parse HEAD' may or may not point to a valid remote.

sding3: `https://${HOSTNAME}/${PATH}/-/tree/${REF}` should work. `${REF}` can be a git SHA, a tag, or a branch.

That's a good point about `git rev-parse HEAD` not being present on the remote since the local update might not have been pushed yet.

prarit: Let me add a --ref option.  I think that is different from a branch so it needs its own option.

zampierilucas (approved):

cmd/project_browse.go
sding3: I tested `https://${HOSTNAME}/${PATH}/-/tree/${REF}` where the `${REF}` was a branch name. I think we should only need the `--ref` option and no longer need the `--branch` option.

prarit: I think we need both -- users could request to see a file on a pushed branch :)

sding3: Wouldn't `--ref` also work for seeing a file on a pushed branch?

sding3: Here are some examples showing branch, tag, and SHA:

|  type | example |
|---|---|
| branch  |   https://gitlab.com/gitlab-org/gitlab-foss/-/blob/doc/test-naming/.gitlab-ci.yml |
| tag | https://gitlab.com/gitlab-org/gitlab-foss/-/blob/v15.11.11/.gitlab-ci.yml |
| SHA | https://gitlab.com/gitlab-org/gitlab-foss/-/blob/55a6d7c9/.gitlab-ci.yml |

PR #871, opened 2023-07-14
Close issue on merge

prarit:


bmeneg:
@prarit, the code rearrangement seems odd. Only `func init()` is left as the final function, the command creation has been the first bit after lib imports and global var declaration, doesn't it?

zampierilucas:
@bmeneg I agree, on newly created function we use `Run: func(...` , so it might be nice to also add it here, since you moving it

cmd/mr_create.go
```diff
@@ -312,6 +242,23 @@ func runMRCreate(cmd *cobra.Command, args []string) {
 		log.Fatal("empty MR message")
 	}
 
+	closeIssues, err := cmd.Flags().GetStringSlice("close-issue")
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	if len(closeIssues) > 0 {
+		closeIssueString := "\n\nCloses "
```
zampierilucas, line 251: What happens if `Closes` is in the first line? Does it still work?
This is the Gitlab regex just for reference `/^\/(?<cmd>close|reopen|...)(?:( |$))(?<arg>[^\/\n]*)(?:\n|$)/`
which if you take all the fuzz, it matches that it is at the start of the line.

PR #872, opened 2023-07-14
cmd/mr_create.go: Fix -f option

prarit:
The -f option was reported to be broken.  This occurred due to a coding error and this simple change will fix it.

Fix the -f option (-F and no options have also been verified to work).


prarit:
@zampierilucas quick fix.  Obviously broken.

zampierilucas (approved):

PR #886, opened 2024-01-10
Update to latest library version

prarit:
Update to the latest library versions.


zampierilucas (approved):

prarit:
@rsteube , there are some caraspace changes here that could use your review.

cmd/issue_create.go
```diff
@@ -181,7 +181,7 @@ func init() {
 			if err != nil {
 				return carapace.ActionMessage(err.Error())
 			}
-			return action.Labels(project).Invoke(c).Filter(c.Parts).ToA()
+			return action.Labels(project).Invoke(c).FilterParts()
```
rsteube, line 184: Woah :astonished:. Calling `FilterArgs` on an `InvokedAction` isn't meant to be possible (at least not without `ToA()`).
Never realized the embedded `Action` is still publc :grin: .

Anyway, what I'm seeing in the diffs should be fine though.

However I've got a couple convenience function now so:
-  `Invoke(c)` isn't needed  anymore in this case.
- [`Action.UniqueList`](https://rsteube.github.io/carapace/carapace/action/uniqueList.html) is also an alternative:
```go
		"label": carapace.ActionCallback(func(c carapace.Context) carapace.Action {
			project, _, err := parseArgsRemoteAndProject(c.Args)
			if err != nil {
				return carapace.ActionMessage(err.Error())
			}
			return action.Labels(project)
		}).UniqueList(","),
```
Though not much a difference here due to the `parseArgsRemoteAndProject`.
There are other cases where this makes it more concise.

PR #888, opened 2024-02-16
Print labels with their colors on lab label list

### LINKED ISSUE #887: show label colors in terminal
would be neat if `lab label  list` could reproduce the true colors in terminal. Dunno if the API returns the info though x)

zampierilucas:
This takes the `label.Color` hex and converts it to color in HexToRGB in utils, then uses is to set the background color.

Closes: https://github.com/zaquestion/lab/issues/887

### DIFF
--- a/cmd/label_list.go
+++ b/cmd/label_list.go
@@ -53,7 +53,10 @@ var labelListCmd = &cobra.Command{
 				description = " - " + label.Description
 			}
 
-			fmt.Printf("%s%s\n", label.Name, description)
+			// Convert hex color to rgb object
+			c := HexToRGB(label.Color)
+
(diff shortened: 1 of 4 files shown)


zampierilucas:
This is what the labels look like with the change.
image

prarit:
I like this change but I think it should be optional.  @zampierilucas I think you should add a config option for this to enable the colors.  That way users can set it in their lab.toml and not worry about it.

zampierilucas:
@prarit I've updated my mr with you suggestion.

bmeneg (approved):
We can make this change generic to every lab command.
Since it's a bigger change, this one is fine wrt the request.

teto:
was the CI failure linked to this PR ? else maybe it's mergeable ?

PR #902, opened 2025-04-04
Add mr list fancy

prarit:
I've always found the project merge_requests page useful.  It shows the
    status of MRs, the CI and Approval statuses, etc.  Let's start doing that
    in the 'mr list' command.  I still like the default simple command, so
    let's add a '--show-state' to get fancy output.
    
    Note, that this requires a per-MR lookup to get the CI status.  That
    slows the output of --show-state down, but it's worth the trade off.
    
    Add --show-state to output similar to the project merge_requests page.
    
### DIFF
--- a/cmd/mr_list.go
+++ b/cmd/mr_list.go
@@ -5,6 +5,7 @@ import (
 	"os"
 	"strconv"
 	"strings"
+	"unicode"
 
 	"github.com/MakeNowJust/heredoc/v2"
 	"github.com/fatih/color"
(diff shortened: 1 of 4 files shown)


JohnVillalovos:
My only concern would be the 3rd party libraries being brought in. How trustworthy are they and the authors?

prarit:
@JohnVillalovos I'm okay with the new packages.  They're all listed on pkg.go.dev and I'll take that as being safe.

JohnVillalovos:
> @JohnVillalovos I'm okay with the new packages. They're all listed on pkg.go.dev and I'll take that as being safe.

Okay. I'm not familiar with that as I haven't done hardly any Go development yet. But if you think it is okay I'll trust you 😊

PR #907, opened 2026-04-11
Update install instructions to use go install

### LINKED ISSUE #904: no new releases have been created since 2022 but installation instructions still state to use pre-built binaries
There appears to have been code modifications as recently as 6 months ago, but the latest release is 3 years old.  If the pre-built release binaries are no longer the best way to get the latest version, I would recommend removing that from the installation instructions in favor of `go install`.

utafrali:
The releases page hasn't been updated since 2022, so pointing users to pre-built binaries isn't super useful. Switched to recommending `go install` instead, which is the standard way to install Go tools anyway.

Fixes #904


utafrali:
closing this out, feel free to reopen if needed

